PR #4, opened 2020-12-29
Automated workflows to publish the docker image

vemonet:
I created 2 different GitHub Actions workflows to build and publish the Docker image

## To DockerHub

I used the official build and push action from DockerHub

https://github.com/marketplace/actions/build-and-push-docker-images

It's a simple workflow that will build and push to `libretranslate/libretranslate:latest` at each push to the `main` branch

Secrets to set:

* `DOCKERHUB_USERNAME`
* `DOCKERHUB_TOKEN` (for security most container registries now use token instead of the password, it should be easy to find where to create it)

## To GitHub Action

I used the workflow provided by GitHub (when going to https://github.com/uav4geo/LibreTranslate/actions/new and clicking on `More continuous integration workflows`). With some customizations

It will publish the container at `ghcr.io/uav4geo/libretranslate` following those rules:

1. Publish a new `latest` image at each push to the `main` branch
2. When you create a new release (e.g. `v1.0.1`), it should publish an image for it with the tag `1.0.1` 
3. The workflow will also run a simple `docker build` for every pull request (to make sure the change did not break the docker image before merging)

> Note that you can easily make it simpler if you don't need all those rules, the workflow is heavily documented and easy to edit, especially if you know your Bash basics 

You don't need to create the image, github will create it auto at the first push as a private image though! You will just need to go to your image and **set it to public after the first push**

You will find the LibreTranslate package here on GitHub website: https://github.com/orgs/uav4geo/packages

Secret to set: 

* `CONTAINER_REGISTRY_GITHUB_TOKEN`, create an access token with `read:packages` and `write:packages` scopes at https://github.com/settings/tokens

Personally I stopped publishing to DockerHub due to the new pull rates which are really restrictives and cause issues in our organization, I am using GitHub Container Registry for all my applications at work and personal and never had any issue, even if it's still `beta`. I just hope they will not set pull rates later :) (but the ongoing GitHub policies to promote open source is quite encouraging)

Here an example of what it looks like when an image is published on GitHub: https://github.com/users/vemonet/packages/container/package/jupyterlab

### DIFF
--- a/.github/workflows/publish-docker-github.yml
+++ b/.github/workflows/publish-docker-github.yml
@@ -1,0 +1,73 @@
+name: Publish to GitHub Container Registry
+# To uav4geo GitHub Container Registry
+# https://github.com/orgs/uav4geo/packages
+
+on:
+  push:
+    # Publish `main` as Docker `latest` image.
+    branches:
+      - main
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
+  # Run tests for any PRs.
+  pull_request:
+  # Enable manual trigger.
+  workflow_dispatch:
+
+env:
+  IMAGE_NAME: libretranslate
+  # The image ID is generated using the GitHub user ID
+  # In our case IMAGE_ID: ghcr.io/uav4geo/libretranslate
+
+jobs:
+  # Test if the docker image build successfully
+  test:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


pierotofy:
This is awesome! Thanks for the PR and the detailed instructions @vemonet  :+1:

PR #7, opened 2021-01-10
Add frontend language default parameter, favicon, move charLimit api, swagger response model

### LINKED ISSUE #6: Add frontend language default parameter
Hello there,
first of all: really great work! 

I would love to see a parameter/config/function to set the default language for the frontend. So that you do not always have to set them by hand.

Here I would like to take over the implementation. But first I would like to clarify a few questions with you:

- does this feature make sense at all?
- Implementation via cookie in the frontend so it remembers for each user?
 OR
- Add two cli arguments and integrate them into the frontend via template variables?
 OR
- cli arguments and fetch in frontend via api. e.g. GET /languages/defaults -> { "source": "en", "target": "es"}
 OR
- cli arguments and fetch in frontend via api. e.g. GET /frontend/settings -> { "source": "en", "target": "es", "darkmode": true}
-> here already in view for  https://github.com/EsmailELBoBDev2/LibreTranslate/commit/68ce70cb78a43b94100ab3eca5ddc7fda9d49d8b

I prefer one of the last two options. But I am not sure if this is a bit overloaded, because the main feature of this project is the translation api and the frontend is more of a showcase? (So this feature should rather be included in the project which uses this API...)?

worldworm:
Hi there.

I'm not very experienced with vue.js, so look twice at the changes. 
**If you find any errors or have any suggestions, tips or requests don't be afraid to let me know.**

- closes #6
- adds favicon
- move charLimit to settings api
- fix swagger response model. It should look like this now:

grafik


pierotofy (approved):
Looks great!

pierotofy:
Thanks @worldworm ! :pray:

PR #12, opened 2021-01-13
add support for auto language

vemonet:
Not perfect, it picks a language not supported sometimes

It could be improved by using the function to get the list of most probable languages, then iterate to get the first we support

I also needed to manually add `auto` to the list of languages after getting it from Argos Translate

And I don't know why but it does not seems the default language have been updated even with this change:

```python
def create_app(char_limit=-1, req_limit=-1, ga_id=None, debug=False, frontend_language_source="auto", frontend_language_target="en"):
```

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -1,6 +1,7 @@
 from flask import Flask, render_template, jsonify, request, abort, send_from_directory
 from flask_swagger import swagger
 from flask_swagger_ui import get_swaggerui_blueprint
+from langdetect import detect
 
 def get_remote_address():
     if request.headers.getlist("X-Forwarded-For"):
@@ -10,7 +11,7 @@ def get_remote_address():
 
     return ip
 
-def create_app(char_limit=-1, req_limit=-1, ga_id=None, debug=False, frontend_language_source="en", frontend_language_target="es"):
+def create_app(char_limit=-1, req_limit=-1, ga_id=None, debug=False, frontend_language_source="auto", frontend_language_target="en"):
     from app.init import boot
     boot()
     
@@ -21,7 +22,7 @@ def create_app(char_limit=-1, req_limit=-1, ga_id=None, debug=False, frontend_la
         app.config['TEMPLATES_AUTO_RELOAD'] = True
 
     # Map userdefined frontend languages to argos language object.
-    frontend_argos_language_source = next(iter([l for l in languages if l.code == frontend_language_source]), None)
+    frontend_argos_language_source = next(iter([l for l in languages if l.code == frontend_language_source or l.code == 'auto']), None)
     frontend_argos_language_target = next(iter([l for l in languages if l.code == frontend_language_target]), None)
     # Raise AttributeError to prevent app startup if user input is not valid.
     if frontend_argos_language_source is None:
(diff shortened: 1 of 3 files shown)


pierotofy:
This is awesome! 

I think the place where to set the default language if you want it to trickle up is https://github.com/uav4geo/LibreTranslate/blob/main/main.py#L19 (in the config arguments).

Perhaps if a language is not supported we could default to English or like you say, iterate through the possibilities.

app/app.py
```diff
@@ -187,9 +188,16 @@ def translate():
...
+        original_source_lang = source_lang
+        if source_lang == 'auto':
+            source_lang = detect(q)
+
         src_lang = next(iter([l for l in languages if l.code == source_lang]), None)
         tgt_lang = next(iter([l for l in languages if l.code == target_lang]), None)
-
+        
+        if src_lang is None and original_source_lang == 'auto':
+            return jsonify({"translatedText": "Detected language not supported (" + source_lang + ")" })
```
worldworm, line 199: Wouldn't it be better to throw an http error code here? So you could also show a matching error message in the frontend or handle it in an application that is using this api.
Otherwise, there is a risk that a user will see this as a translation because a successful 200 is coming.

app/templates/index.html
```diff
@@ -299,6 +299,7 @@ <h5 class="white-text">LibreTranslate</h5>
 		  if (this.status >= 200 && this.status < 400) {
 			// Success!
 			self.langs = JSON.parse(this.response);
+			self.langs.push({ name: 'Auto (experimental)', code: 'auto' })
```
worldworm, line 302: This makes the auto option selectable only in the frontend. The api GET /languages does not contain the auto option.

worldworm (changes requested):
Hi there.
First of all, good work. I find an auto function very useful!

I have added a few suggestions for improvement here. Some more or less useful. But maybe they help.

Also a new API interface would be very helpful to return only the recognized language. (But maybe this is more a suggestion for a later feature).
**BUT**
Alternatively here the consideration whether it would be easier/cleaner in the frontend when selecting auto that only a language detection api is used to recognize. Then the frontend automatically change the language to the new recognized language and only then triggers the translation api with the correctly selected source language (this could reduce the load on the translation api, since in the event of an error, it is not necessary to translate several times.). This could make it a bit easier in the frontend and error consistent as well but adds an additional api request.
Also the user gets a feedback which language was recognized at all. There could be an addititonal "confidence level" in the detection api. The english fallback @pierotofy talked about would then be not part of the translation api and could be better implementend in the frontend (shipped via the new settings api) or the software that is using this api?

app/app.py
```diff
@@ -187,9 +188,16 @@ def translate():
         if char_limit != -1:
             q = q[:char_limit]
 
+        original_source_lang = source_lang
```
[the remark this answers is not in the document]
vemonet: Needed once later to check if it is originally `auto` (allows me to keep `source_lang` as it is without too much changes)

app/app.py
vemonet: No, because LibreTranslate try to translate at every key push, so you are getting a lot of languages that are not supported before finishing typing. It was actually the main issue to take of with the UI. But this should be improved by asking for the list of most probable langs and using the first that we support

```
from langdetect import detect_langs
>>> detect_langs("Otec matka syn.")
[sk:0.572770823327, pl:0.292872522702, cs:0.134356653968]
```

app/templates/index.html
vemonet: Yep, the issue is that LibreTranslate use the list loaded by Argos, and I don't want to change this Argos list, so we need to add `auto` manually in the UI and API lists (got to do the last one, thanks for the notice!)

pierotofy:
Overall this is a good start, thanks @vemonet !

pierotofy:
It's kind of fun to see what the language guesses are. I feel like it the library misses a lot of simple cases. For example typing `hello` returns `[fi:0.5714244460768835, nl:0.2857151445973061, en:0.14285635995117174]` :smile: 

Definitely room for improvement.

pierotofy:
I've pushed some changes in `main` to consider the list of probabilities for selecting a language via auto. :+1:

PR #22, opened 2021-01-19
Add support for batch translation

### LINKED ISSUE #21: Batch translation
It would be very nice if the `translate` endpoint would accept multiple texts (or `q`s) at once, like DeepL does.

This should significantly improve the translation performance.

* If `q` is a string, translate it, and return the translation as a string.
* If `q` is an array (list of strings), translate each item, and return the translations as an array (list of strings).

Note that DeepL has a limit of 50 texts per requests, which could be an argument like `--char-limit` (e.g. `--text-limit` or `--q-limit`).

grafik
(Source: https://www.deepl.com/docs-api/translating-text/request/)

caugner:
Fixes #21

Note that this includes a breaking change w.r.t. the `--char-limit`:
* So far, the input string (`q`) was simply trimmed to the character limit.
* Now, we abort with a 400 error if the request exceeds the character limit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ docker-compose up -d --build
 | --port        | Set port to bind the server to | `5000`               |
 | --char-limit        | Set character limit | `No limit`               |
 | --req-limit        | Set maximum number of requests per minute per client | `No limit`               |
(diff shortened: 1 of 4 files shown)


pierotofy:
This is awesome. Thanks @caugner !

PR #31, opened 2021-01-28
Add frontend translation timeout parameter

### LINKED ISSUE #29: Avoid the error "Slowdown: 20 per 1 minute"
I propose to avoid this by waiting 1 second to translate the text.

That is, when the user stops typing for 1 second, perform the translation automatically.

worldworm:
Closes #29 

I included the value via the template syntax.
In addition, it can also be accessed via the settings api, if someone want to use this value in separate applications to be able to change it centrally later.

Feedback and suggestions for improvement are always welcome!


pierotofy:
Looks amazing! :1st_place_medal: 

Thanks @worldworm !

PR #39, opened 2021-02-10
Add badges to the readme to display tests results and python versions

vemonet:
Add badges to the readme to display tests results and python versions from pypi, makes it much easier to catch when the tests start to fail


pierotofy:
Fantastic, thanks!

PR #42, opened 2021-02-10
Add offline parameter

Trekky12:
This PR includes a command line parameter to enable offline usage after initializing and downloading models.


pierotofy:
Thanks @Trekky12 ! Very useful.

PR #45, opened 2021-02-18
fix static package_data for offline mode

ErQrYfkrju:
package_data 'static/*'  won`t include sub directories so offline mode would result in 404 for static resources under js, css, fonts


pierotofy:
Thanks for the fix @ErQrYfkrju ! :pray:

PR #59, opened 2021-03-11
Include language models into Docker image

mammo0:
Hello,

normally an internet connection is required on the first run of the Docker image to download the language models. But if you got a complete offline Docker environment then you can't start the container.

So I modified the Dockerfile and added a build argument `with_models` (default: `false`). If it's set to `true`, the language models get downloaded during the image creation process. Afterwards the image is ready to go in an offline Docker environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,20 @@
 FROM python:3.8
 
+ARG with_models=false
+
 WORKDIR /app
 
 RUN pip install --upgrade pip
 
 COPY . .
+
+# check for offline build
+RUN if [ "$with_models" = "true" ]; then  \
+        # install only the dependencies first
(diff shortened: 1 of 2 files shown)


pierotofy:
Love it, thanks @mammo0 !

mauro-gobbi:
Hello, can we achieve the same result if not using Docker? I mean, in case we use `pip install -e .` is there a way to download at once all the required data and then being able to run **libre translate** from a completely offline (virtual) machine?
Thanks!
Mauro

mammo0:
Have you tried running `./install_models.py` after installing the dependencies with `pip install -e .`?
It's the next step that is also done in the Dockerfile.

This script should download the required language models.

PR #60, opened 2021-03-11
use polyglot for language detection and transliteration

mammo0:
Hello,

I noticed that the if you want to translate from Russian to English, argostranslate expects the input string to be in Cyrillic characters. If you use a Russian text in Latin characters, no translation is done.

So I searched for a Python library that can convert Latin to Cyrillic characters and found [polyglot](https://pypi.org/project/polyglot/). In addition to transliteration this library also supports language detection.

So first I replaced the current language detection implementation. Then I added support for transliteration.

I've tested my changes in an Docker instance.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,2 +1,10 @@
+import os
+from appdirs import user_data_dir
+
+# override polyglot path
+import polyglot
+polyglot.polyglot_path = os.path.join(user_data_dir(appname="LibreTranslate", appauthor="uav4geo"), "polyglot_data")
+
+
 from .main import main
 from .manage import manage
--- a/app/app.py
+++ b/app/app.py
@@ -2,13 +2,9 @@ import os
 from flask import Flask, render_template, jsonify, request, abort, send_from_directory
 from flask_swagger import swagger
 from flask_swagger_ui import get_swaggerui_blueprint
-from langdetect import detect_langs
-from langdetect import DetectorFactory
 from pkg_resources import resource_filename
 from .api_keys import Database
 from app.language import detect_languages
-
-DetectorFactory.seed = 0 # deterministic
 
 api_keys_db = None
 
--- a/app/app.py
+++ b/app/app.py
@@ -6,6 +6,7 @@ from langdetect import detect_langs
 from langdetect import DetectorFactory
 from pkg_resources import resource_filename
 from .api_keys import Database
+from app.language import detect_languages
 
 DetectorFactory.seed = 0 # deterministic
 
@@ -56,11 +57,6 @@ def create_app(args):
 
     from app.language import languages
     app = Flask(__name__)
-
-    # For faster access
-    language_map = {}
-    for l in languages:
-        language_map[l.code] = l.name
 
     if args.debug:
         app.config['TEMPLATES_AUTO_RELOAD'] = True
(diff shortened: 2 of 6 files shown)


pierotofy:
This is awesome @mammo0 ! :star: 

Language detection seems to work a lot better with polyglot. Really impressive.

Thanks! Code looks great.

mammo0:
I'm glad I could helped.

PJ-Finlay:
TIL: https://en.wikipedia.org/wiki/Romanization_of_Russian

pierotofy:
I didn't know that was a thing either. Super interesting.

schrmh:
@pierotofy Maybe things have changed but polyglot sucks at detecting german while langdetect is pretty reliable (nl is close to de so it's ok it failed once):
image
image

Looks like a bunch of things are detected with 0% confidence that it is english
image
However, it thinks with a confidence of 88% that "Sprache" is english and spills out a interesting japanese translation for that...:
image

Also what happened here? "watashi" is something every japanese beginner knows, why does it translate it to "Home" that isn't even a german word...
image
And another thing to note here: Why is the confidence only 90%? I wish it would respect the models I have installed (german, english and japanese) thus it should display a number even more close to 100% since german and english use a different alphabet.

I just seriously just can't use LibreTranslate if it fails at simple things like this. It is just to unreliable when polyglot is used. Please switch back to langdetect or combine both somehow (e.g. when polyglot has a low confidence then use langdetect if it is the case that langdetect is bad detecting other languages...) or give the user the option to choose which to use.

pierotofy:
We'd welcome contributions to improve language detection.

schrmh:
Looks like Issues don't accept a feature request and bad detection is not a bug, is it? So...
To add something more to why to get rid of polyglot again (or to make it configurable):
https://github.com/pemistahl/lingua-py contains accuracy graphs (**I picked only two, you should really check out the graphs there cause there are even some for individual languages**):
image
image

polyglot uses a library which depends on cld2 (https://polyglot.readthedocs.io/en/latest/Detection.html). Those images show that langdetect, which was once used by LibreTranslate, is more accurate. However, there is one better option according to graphs and that is lingua itself, so it might be worth checking that out.

It does looks really promising from a first look and the README provides a bunch of examples, even showing how to only care about a list of languages you want to detect (like in my case).

I tried it with all words I mentioned and I included all languages that polyglot and langdetect claimed to detect (except for Manx since lingua can't detect that language that isn't even spoken by 2000 people) and it is pretty good:
image
While it did not detect "Niemand", it is completely fair since that is indeed a word used in afrikaans and dutch (https://en.wiktionary.org/wiki/niemand), the confidence values look pretty good as well:
```
[ConfidenceValue(language=Language.AFRIKAANS, value=0.3896039149451188),   
ConfidenceValue(language=Language.DUTCH, value=0.33884170298425187),   
ConfidenceValue(language=Language.GERMAN, value=0.2244259347646024),   
ConfidenceValue(language=Language.ENGLISH, value=0.0471284473060269),   
ConfidenceValue(language=Language.JAPANESE, value=0.0)]   
```
(in contrast, langdetect did not detect "Neuland" as german at all, iirc about 14% or so was tagalog and the rest dutch).
When I only include german, english and japanese it does of course detect "Niemand" as german.

pierotofy:
Thanks, the library looks very promising. Perhaps I'll find some time to replace cld2 (we no longer use polyglot, we use cld2 directly).

mammo0:
> we no longer use polyglot, we use cld2 directly

@schrmh you can have a look at the current implementation at https://github.com/LibreTranslate/LibreTranslate/blob/main/libretranslate/detect.py

So my PR is already obsolete. Maybe the progress of implementing the **lingua** library can be tracked in a separate issue. Or you can open a PR by yourself :)

PR #67, opened 2021-03-28
Add option --load-only to set available languages

### LINKED ISSUE #51: [FR] Limit to two languages
It crashed my 1GB vm when I scrolled through the dropdown languages, I assume it had to load all the language models and process them at once. The API already supports `/languages` but I was unable to find anywhere to limit to a single model. Thanks!

Philippe-Cholet:
Add an option `--load-only` to the main parser, we give comma-separated language codes and it filters available packages from argostranslate to not install all packages but only the ones the user actually need.
The variable `load_only` will be a list of strings (language codes), or `None` by default in which case we do not filter available packages.
I also check that given language codes are available and that we did not discard all packages. Both would raise a ValueError.
It should close #51 
Any modification is welcome.


pierotofy (approved):

pierotofy:
Looks great! Thanks @Philippe-Cholet ! :+1:

PR #68, opened 2021-03-28
Fix `languages` reload

StephenSorriaux:
As this object can be updated from outside its own module, the module should be imported and
used instead.

Relates to https://github.com/uav4geo/LibreTranslate/issues/54


pierotofy:
Thanks @StephenSorriaux ! Looks great. :clinking_glasses:

PR #71, opened 2021-03-31
LibreJS support

jdu9:
Now LibreTranslate works with LibreJS. :)

See https://www.gnu.org/software/librejs/free-your-javascript.html for future reference.


pierotofy:
Thanks!

Issue #82, opened 2021-05-11
Flood protection


pierotofy:
Welp, that time has arrived where some users have started to abuse the public API at https://libretranslate.com, despite asking people not to do so. This was expected, and probably a good sign :smile: 

The app needs a flood protection system to ban offenders.

I've temporarely disabled translations on the website, unless a user has setup an API key. I'll bring them back once flood protection is in place.

PJ-Finlay:
Also adding some of the mirrors to the README is probably a good idea to direct people that want to regularly use LibreTranslate away from libretranslate.com. They would probably appreciate the traffic too.

- https://libretranslate.de/
- https://translate.astian.org/

pierotofy:
Updated libretranslate.com, so people can now translate again using moderate request limits and repeat offenders will be banned for a month.

PR #87, opened 2021-05-16
Flood protection

pierotofy:
Adds additional daily requests limits on top of the request limits per minute. This should discourage improper use, but still allow people to translate a few sentences in moderation.


pierotofy:
Closes #82

PR #89, opened 2021-05-17
Programmatic api keys access

pierotofy:
Flooding seems to continue even after flooding protection has been introduced. I suspect this is due to somebody having added a client side application in some app, somewhere, causing many IP addresses to flood the API. This is obviously not sustainable. 

I've added another layer of checking, whereby we allow the use of API calls to be made from an origin (e.g. libretranslate.com, but not from other clients). This is a simple check and can be spoofed, but should at least ﻿work for less determined offenders.


pierotofy:
Legitimate users can now also enjoy a bit higher limits as a result of this. :+1:

PR #95, opened 2021-06-02
Add html5 structure + improve UX/UI

idotj:
- Added a valid HTML5 structure tags (such as `<header>` and `<main>`)
- Removed css inline styles
- Fix html indentation
- Restyle language selectors
- Improve textarea styles
- Improve erase text + copy translated text buttons
- Relocate footer slogan


pierotofy:
Thanks @idotj ! This looks like a great UI improvement. Seems like you have a very good eye for design. :)

PR #98, opened 2021-06-12
Features + fix bugs (frontend)

idotj:
**Features:**
- Change translated textarea to "readonly"
- Add an associated label to each textarea (Accessibility WCAG 2.1)
- Background, foreground and icon colors do not have a sufficient contrast ratio (Accessibility WCAG 2.1)
- Add rel="noopener noreferrer" to all external links

**Bugs:**
- When a translated text is deleted (by the "x" button) the Response text area keeps the JSON code on screen
- Duplicated id in html: "index-banner"
- Element <div> not allowed as child of element <label> (loadingTranslation progress bar)
- VueJS has a wrong call to an external variable (needs simple quote)


pierotofy:
Thanks @idotj ! Looks fantastic.

PR #99, opened 2021-06-12
Set selected attributes

pierotofy:
Sets the "selected" attribute when choosing a source/target language.﻿


pierotofy:
cc @idotj

PR #103, opened 2021-06-13
Add downArrow to select language + fix select language problem with small screens

idotj:
- Add down arrow to language selector + manage dynamic width depending on the option selected.
- Fix problem with language selector in small screen resolutions


pierotofy (approved):

pierotofy:
Thanks @idotj ! :+1:

PR #109, opened 2021-07-04
update pyicu  to fix build on macos

farfromrefug:
Right now brew installs icu4c 69.x which pyicu only supports from 2.7


pierotofy:
Thanks @farfromrefug !

PJ-Finlay:
Is this related? https://community.libretranslate.com/t/pyicu-fails-to-install-on-ubuntu-20-04/23

farfromrefug:
@PJ-Finlay no the error in there is fixed by the answer there. You need to install icu first (icu4c on macos)

PR #112, opened 2021-07-08
Add min-height to main app

Dafnik:
Adds min-height of 80vh to `<main id="app">`

The footer will stay always at the bottom on big screens and looks normal on mobile devices.

Before:
image

After:
image

Mobile after:
image


pierotofy:
Thanks @Dafnik ! Looks great.

PR #115, opened 2021-07-12
Add env support

euberdeveloper:
In this pull request, I added an env vars layer between the cli default values and the passed args.

For instance, the default value of `HOST` is the old one (`127.0.0.1`), but if I add an env var `LT_HOST=www.ciao.com/api`, the `host` default value will be that one. If I pass both that env var and an argument `--host www.google.it/api`, the arg one wins.

This was done for each option, maintaining all the old values, this means that it is not a breakable change (unless someone has an env var with he same name)

The env var has the same uppercase-snake-case name of the cli option, but with a prefix `LT_` prepended to it (`req-flood-threshold` -> `LT_REQ_FLOOD_THRESHOLD`)

This could be for instance used on a docker-compose env file

Signed-off-by: euberdeveloper <euberdeveloper@gmail.com>

### DIFF
--- a/app/default_values.py
+++ b/app/default_values.py
@@ -1,0 +1,119 @@
+import os
+
+_prefix = 'LT_'
+
+def _get_value_str(name, default_value):
+    env_value = os.environ.get(name)
+    return default_value if env_value is None else env_value
+
+def _get_value_int(name, default_value):
+    try:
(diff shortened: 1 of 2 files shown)


pierotofy (approved):
Looks great!! Thanks @euberdeveloper :heart:

PR #118, opened 2021-07-25
fix(default_values.py): remove debug print

euberdeveloper:


euberdeveloper:
Sorry, I noticed it just now by using it

pierotofy:
Nice catch! :+1:  Thanks.

PR #121, opened 2021-07-30
Mention Argos in leading paragraph

jensli:
One of the first things that people will probably want to know about this
project is how the actual translation works. Therefore, add mention of
Argo to the leading paragraph of the readme.


pierotofy (approved):
:+1: Thanks!

PR #128, opened 2021-09-04
Add dir="auto" to fix rtl lang problem

alefvanoon:
After:
image
Before:
image

Computers support two kinds of text flow:

* "LTR": from left to right, like most western languages
* "RTL": from right to left, like Arabic, Persian and Hebrew


PJ-Finlay:
Nice looks good!

pierotofy:
Thanks!!

PR #135, opened 2021-09-11
Feature/add html translation

dingedi:
It should work


PJ-Finlay:
Looks good to me

pierotofy:
Looks great @dingedi ! :pray: 

Thank you so much! I've added a note in the README for the new parameter and tweaked slightly the swagger definition.

PR #151, opened 2021-10-08
[WIP] Suggesting corrections to translations

dingedi:
IN PROGRESS.

I started working on it, I open the PR now if you have any suggestions.
I'm not a very good designer, so maybe the design needs to be improved.

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -566,6 +566,9 @@ def create_app(args):
                 frontendTimeout:
                   type: integer
                   description: Frontend translation timeout
+                suggestions:
+                  type: boolean
+                  description: Whether submitting suggestions is enabled.
                 language:
                   type: object
                   properties:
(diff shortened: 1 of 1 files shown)


PJ-Finlay:
Looks good to me!

This probably should be disabled by default and run if the server operator enables it with a command line flag. That way we're only prompting users to correct translations on servers that are likely to contribute the generated data.

image


image

pierotofy:
Working on adding a flag to enable/disable suggestions! The UI looks great @dingedi

pierotofy:
Added the swagger docs :+1: 

Good point about the privacy page; Google Translate does not have anything linked directly on their translation page when you submit contributions, although I'm sure they have a note somewhere buried in one of their policies.

pierotofy:
Agreed.

PR #152, opened 2021-10-09
Fix display issue on libretranslate.com

dingedi:
The css file was not updated despite having cleared my cache several times. The problem came from "?t=2" after the css file, I don't know if it had any other utility?

I added a parameter with the version number

image

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -1,6 +1,7 @@
 import os
 from functools import wraps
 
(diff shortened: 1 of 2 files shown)


pierotofy:
Thanks @dingedi ! The parameter was there just to invalidate the cache, but it had to be updated manually (your approach is better)

PR #162, opened 2021-10-30
Expose require_api_key_origin in frontend settings

fushinari:
This is useful for applications that want to know beforehand if an API key is necessary before accessing the running instance.

This is necessary for me as a part of adding proper support for LibreTranslate API keys in [Dialect](https://github.com/dialect-app/dialect).

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -756,6 +756,9 @@ def create_app(args):
                 frontendTimeout:
                   type: integer
                   description: Frontend translation timeout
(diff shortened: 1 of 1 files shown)


fushinari:
`keysEnabled` and `keyRequired` could be separate too. What would be preferable?

pierotofy (approved):

pierotofy:
Thanks @fushinari !

PR #164, opened 2021-10-31
Require API key for submitting suggestions

fushinari:
Also removes limiter exemption

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -811,7 +811,7 @@ def create_app(args):
         )
 
     @app.route("/suggest", methods=["POST"])
-    @limiter.exempt
+    @access_check
     def suggest():
         """
         Submit a suggestion to improve a translation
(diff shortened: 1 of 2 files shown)


pierotofy:
This would break the current client code, as https://github.com/LibreTranslate/LibreTranslate/blob/main/app/static/js/app.js#L288 does not forward the API key.

fushinari:
Oops, sorry, will fix

fushinari:
@pierotofy I'm sorry but I can't really test my changes unfortunately, because I get an error when trying to setup LibreTranslate locally:
```
ERROR: Could not find a version that satisfies the requirement ctranslate2==2.4.0 (from versions: 2.6.0)
ERROR: No matching distribution found for ctranslate2==2.4.0
```

pierotofy:
No worries, looks good! Thanks. :pray:

PR #165, opened 2021-11-03
Expose api_keys in frontend settings

fushinari:
This is useful for applications (like Dialect) that want to know beforehand if usage of an API key is supported.

Sorry, I know #162 should have covered this use case. Unfortunately, at the time, I wasn't yet sure of our requirements. But now, I realize that we should probably allow users to use API keys even if it's not necessary but simply supported.


pierotofy:
Thanks @fushinari :+1:

PR #169, opened 2021-11-05
decrease counter if the user wait after the slowdown notice

dingedi:
@pierotofy it should work


pierotofy (approved):
Thanks @dingedi ! Looks great.

pierotofy:
Deployed on libretranslate.com.

PR #173, opened 2021-11-09
Reduce Docker image size

### LINKED ISSUE #170: Reduce Docker image size
Currently the Docker image is based on `python:3.8` which finally points to `3.8.12-bullseye`. There are also `3.8-slim` and `3.8-alpine`. Here is a comparison of these base image sizes:

- `3.8.12-bullseye`: 331 MB
- `3.8-slim`: 44 MB
- `3.8-alpine`: 16 MB

Is there a specific reason to go with the largest option `3.8.12-bullseye`? If not, then the switch to `3.8-slim` should be seamless. The switch to `3.8-alpine` would require some effort. But it shouldn't be too much.

Would you generally be open to a switch? From my point of view there shouldn't be any disadvantages.

thomas-mc-work:
Fixes #170

Size is reduced from 4.36 GB to 2.83 GB.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,4 @@
 .git*
+Dockerfile
+.Dockerfile.swp
+LICENSE
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,14 @@
-FROM python:3.8
+FROM python:3.8.12-slim-bullseye
 
 ARG with_models=false
 
 WORKDIR /app
+
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update -qq \
+  && apt-get -qqq install --no-install-recommends -y libicu-dev pkg-config gcc g++ \
+  && apt-get clean \
+  && rm -rf /var/lib/apt
 
 RUN pip install --upgrade pip
 
@@ -17,7 +23,8 @@ RUN if [ "$with_models" = "true" ]; then  \
     fi
 
 # Install package from source code
-RUN pip install .
+RUN pip install . \
+  && pip cache purge
 
 EXPOSE 5000
-ENTRYPOINT [ "libretranslate", "--host", "0.0.0.0" ]
+ENTRYPOINT [ "/home/worker/.local/bin/libretranslate", "--host", "0.0.0.0" ]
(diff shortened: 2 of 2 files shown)


pierotofy:
Thanks @thomas-mc-work :pray: !

thomas-mc-work:
You're welcome!

BTW: I wasn't successful with the switch to alpine. The `pip install .` failed at building the package `sip`. I couldn't figure out the final reason.

thomas-mc-work:
Here is my approach, for anybody carrying on:

    FROM python:3.8.12-alpine

    ARG with_models=false

    WORKDIR /app

    RUN pip install --upgrade pip
    RUN apk add --no-cache icu-dev libxml2-dev libxslt-dev gcc musl-dev

    COPY . .

    # check for offline build
    RUN if [ "$with_models" = "true" ]; then  \
            # install only the dependencies first
            pip install -e .;  \
            # initialize the language models
            ./install_models.py;  \
        fi

    # Install package from source code
    RUN pip install . \
      && pip cache purge

    EXPOSE 5000
    ENTRYPOINT [ "libretranslate", "--host", "0.0.0.0" ]

PJ-Finlay:
I think sip is part of PyQt for the Argos Translate GUI. It's possible to install Argos Translate without PyQt but probably not worth it in this case. Future versions of Argos Translate should probably have the GUI in a separate repo/package for this reason.

thomas-mc-work:
> but probably not worth it in this case

Why not? Do you know how to exclude it? I was wondering anyway why PyQt would be required at all.

dingedi:
Yes good idea to separate argos-translate and the GUI with PyQT, this will reduce the size of all server installations and libretranslate.

nettnikl:
Hey @PJ-Finlay, i was wondering if you need help with the separation of argos-translate-gui? Not sure if i can assist here, but i'm interested in getting a LibreTranslate build without PyQT working, and was kindly forwarded to this issue.

PJ-Finlay:
[argos-translate-gui](https://github.com/argosopentech/argos-translate-gui) live in `1.6.1`.

The plan was to wait to do this with other breaking changes but it looks like there's value in doing it now so I did.

Let me know if you have any issues.

PR #182, opened 2021-12-09
Properly escape data in the request code

qgustavor:
It was only escaping the first quote, all other quotes and other characters that require to be escaped (like line breaks) were not being escaped. JSON.stringify is a good function to handle this.

Calling `this.$options.filters.escape` looks weird to me, I would just call `JSON.stringify` directly or, if a more complex escape logic was needed, move this function to a method so it can be called by `this.escape`. By the way, filters were removed in Vue 3. I kept the current `filter` call, but consider changing it in future.


pierotofy (approved):

pierotofy:
Nice catch, thanks @qgustavor ! :+1:

PR #190, opened 2021-12-23
Upgrade to Argos Translate 1.6.1

dingedi:
Remove PyQt5


pierotofy:
Nice! This will make it easier to port to architectures that don't have QT binaries.

PR #191, opened 2021-12-31
chore: update url to avoid redirect

SethFalco:
Since this repository lives under the LibreTranslate organization now, figured I'd update the docs to point to it.


pierotofy:
Ups! Good catch :+1:

PR #196, opened 2022-01-06
README: remove non-working instances

fushinari:
Remove instances that don't work anymore


pierotofy:
I suppose we could re-add it.

PJ-Finlay:
I agree, I think it's one of the most widely used instances. Searching "LibreTranslate" on DuckDuckGo libretranslate.de is one of the top results. The fact that it's widely used is probably why it frequently goes down. Leaving it on people have more information/options, but leaving it off could direct traffic away from it which could be a good thing if they're at capacity.

pierotofy:
Done! https://github.com/LibreTranslate/LibreTranslate/commit/27d076a1b467c93cb1f6e3aaf4fcf719c8e2d6f3

PR #198, opened 2022-01-17
Closes #197

### LINKED ISSUE #197: Color contrast fail under AA WCAG standards (LibreTranslate website)
Checking the colors with a tool to validate the AA level WCAG will fail because the contrast doesn't 

After checking the website with a WCAG Color contrast validator, I saw that the recommended level (AA) fails because of the low contrast between the light blue and white:

image

A change of colors will fix the problem.

idotj:
Improve color contrast for WCAG standards + add underline to links

### DIFF
--- a/app/static/css/main.css
+++ b/app/static/css/main.css
@@ -7,6 +7,10 @@ select {
   font-family: Arial, Helvetica, sans-serif !important;
 }
 
+a {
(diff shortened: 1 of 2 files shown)


pierotofy:
Thanks @idotj :pray: Looks great!

Issue #203, opened 2022-01-31
HTML Entity codes are being displayed raw


Eboubaker:
Here result should be: `entité n'a pas été supprimée` but the apostrophe entity code is being displayed as raw html

![image](https://user-images.githubusercontent.com/37766821/151781527-ff5e9563-3544-46dd-b78b-f64c4ea0c57f.png)
Input : `entity was not Removed` with big `R`  letter

dingedi:
after checking the bug is not always present, would it not be a bug of argos rather? @PJ-Finlay 

![image](https://user-images.githubusercontent.com/64697405/151782777-73454f49-eab6-496a-95e3-3fec7873d3c1.png)

PJ-Finlay:
The issue is that the [CTranslate2](https://github.com/OpenNMT/CTranslate2) seq2seq sometimes generates these entity codes because there are entity codes in the training data.

Previously when we had this issue I added a filter in the training scripts to remove them but removed it because it was slow. I'm hoping that by using more and better data the model is able to figure out not to do this on its own. If this becomes a common issue we can try adding more data with apostrophes and other HTML entities being translated correctly as Unicode.

PJ-Finlay:
https://github.com/argosopentech/argos-translate/issues/39

Eboubaker:
Is it possible if you add the temporary filter in the frontend, so it won't affect the backend.

PJ-Finlay:
A front end filter is possible. The issue with filtering them from the data is that searching millions of lines took a long time and was slowing down training.

dingedi:
I'll try to take care of it this week if I have time. It will indeed fix this bug for libretranslate but it will still be present in argos.
would it be interesting to add the filter to the output of the argos translation @PJ-Finlay ?

PJ-Finlay:
For Argos Translate I want to stick with the strategy of fixing this with better models but I think this filtering would be valuable to have in LibreTranslate.

dingedi:
> For Argos Translate I want to stick with the strategy of fixing this with better models but I think this filtering would be valuable to have in LibreTranslate.

Okay yes i think its better than adding fix for each translations bug

PR #204, opened 2022-02-06
unescape html entity codes (ex: &apos)

dingedi:
Fix [https://github.com/LibreTranslate/LibreTranslate/issues/203](https://github.com/LibreTranslate/LibreTranslate/issues/203)

I checked, normally this does not cause a bug for the html translation.


pierotofy (approved):
Looks good, thanks!

Issue #185, opened 2021-12-17
en as frontend source language is not supported


lizongshen:
[root@DESKTOP-B0B9UFO code]# docker run -ti --rm -p 5000:5000 libretranslate/libretranslate
Updating language models
ERROR:root:(RemoteDisconnected('Remote end closed connection without response'),)
Cannot update models (normal if you're offline): Local package index not found, use package.update_package_index() to load it
INFO:apscheduler.scheduler:Adding job tentatively -- it will be properly scheduled when the scheduler starts
INFO:apscheduler.scheduler:Added job "remove_translated_files" to job store "default"
INFO:apscheduler.scheduler:Scheduler started
Traceback (most recent call last):
  File "/usr/local/bin/libretranslate", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/site-packages/app/main.py", line 113, in main
    app = create_app(args)
  File "/usr/local/lib/python3.8/site-packages/app/app.py", line 135, in create_app
    raise AttributeError(
AttributeError: en as frontend source language is not supported.
INFO:apscheduler.scheduler:Scheduler has been shut down

Medallyon:
I get the same `AttributeError`, with a different stacktrace:

```bash
medallyon@MEDALLYON:/mnt/d/Medallyon/Desktop/transplex$ libretranslate
Updating language models
Found 44 models
Downloading Arabic → English (1.0) ...
ERROR:root:(URLError(gaierror(-5, 'No address associated with hostname')),)
Cannot update models (normal if you're offline): expected str, bytes or os.PathLike object, not NoneType
Traceback (most recent call last):
  File "/home/medallyon/.local/bin/libretranslate", line 8, in <module>
    sys.exit(main())
  File "/home/medallyon/.local/lib/python3.8/site-packages/app/main.py", line 107, in main
    app = create_app(args)
  File "/home/medallyon/.local/lib/python3.8/site-packages/app/app.py", line 97, in create_app
    raise AttributeError(
AttributeError: en as frontend source language is not supported.
```

This is on Ubuntu 20.04 with the Linux Subsystem on Windows 10. I can't install it on Windows 10 because some of the dependencies (pyICU, polyglot) don't want to install.

PJ-Finlay:
In both cases the underlying issue is that the models failed to download.

Ex:
```
ERROR:root:(RemoteDisconnected('Remote end closed connection without response'),)
```

```
[...]

PR #206, opened 2022-02-07
set default language instead of raise error

dingedi:
Close [https://github.com/LibreTranslate/LibreTranslate/issues/185](https://github.com/LibreTranslate/LibreTranslate/issues/185)


pierotofy (commented):
Thanks!

Issue #193, opened 2022-01-04
Improvements for the LibreTranslate Website


SethFalco:
The website has some issues on mobile, I'd be happy to fix it myself, but just wanted to keep track of it here.
There are also some other minor improvements I'd like to suggest.

* [x] The `Translate Text` and `Translate File` buttons break on small resolutions.
* [ ] The "LibreTranslate" should be hidden, and the logo should anchor right on small resolutions.
* [ ] The text boxes should fall under each other, similar to the request/response boxes on small resolutions. (The horizontal arrows should rotate 90deg.)
* [x] Hide the "Download" text, and just resize the Download button to the icon on small resolutions.
* [x] Use query parameters to automatically select input/output language when. (ie convenient access to `Detect Language` → `English`)

I'll submit PRs for these over time, just doing them one by one.

pierotofy:
Thanks @SethFalco ! These are most welcomed improvements.

PR #209, opened 2022-02-07
keep selected languages and text in url

dingedi:
Last task in [https://github.com/LibreTranslate/LibreTranslate/issues/193](https://github.com/LibreTranslate/LibreTranslate/issues/193)

add in query url `source, target and q`

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -82,6 +82,24 @@ document.addEventListener('DOMContentLoaded', function(){
                         return;
                     }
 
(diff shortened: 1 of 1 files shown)


pierotofy (approved):
Cool! I think this might run into some issues if the text exceeds 2048 characters, https://stackoverflow.com/questions/812925/what-is-the-maximum-possible-length-of-a-query-string#:~:text=Although%20officially%20there%20is%20no,a%20max%20of%202048%20characters. but we'll see.

PR #215, opened 2022-02-20
app: Fail when giving invalid API keys

fushinari:
When an API key is passed, fail in the case of an invalid API key even if an API key is not required. This allows the user to know that the API key is invalid. Otherwise, they work under the assumption that the API key is correct, even though it is not.


pierotofy (approved):
Thanks @fushinari :+1: looks great!

PR #216, opened 2022-02-20
suggestions: Abort when missing parameters

fushinari:
Earlier, it would create an `sqlite3.IntegrityError` on the server side instead.


pierotofy (approved):
Good addition, thanks!

PR #218, opened 2022-03-04
Patch requirements.txt

ZenulAbidin:
Currently the `docker build`, `pip install -r requirements.txt` and `python run main.py` commands all fail with the current requirements.txt due to deprecated or removed APIs, so I pinned older versions of nested dependencies inside the file.

In particular:

- A new version of `itsdangerous` is being pulled which [removed](https://serverfault.com/questions/1094062/error-from-itsdangerous-import-json-as-json-importerror-cannot-import-name-j) the `json` submodule. This causes runtime failure and breaks Docker build
- `argostranslate` and `ctranslate2` had to be downgraded to an older version and inserted respectively for the pip install process to succeed.


pierotofy:
Hey @ZenulAbidin :hand: thanks for the patch, but I'm not sure I understand the need to downgrade argostranslate, the current version built and pushed correctly to docker hub just a few days ago?

ZenulAbidin:
@pierotofy I downgraded argostranslate just in case it caused dependency errors with ctranslate2, I will try again with the original argostranslate version and report the results.

ZenulAbidin:
@pierotofy Confirmed. It works with argostranslate 1.6.1, the original version.

dingedi:
I think you should not specify a version for ctranslate2 because it is specified by argostranslate, this will avoid problems.

requirements.txt
```diff
@@ -1,6 +1,6 @@
-ctranslate2==1.16.2
+ctranslate2==2.10.1
```
dingedi, line 1: I think you should not specify a version for ctranslate2 because it is specified by argostranslate, this will avoid problems.

ZenulAbidin: I inserted this line before updating `pip`, I'll go check if it works without it on the newer version of pip.

It's possible that this PR may not even be necessary if `pip install` works without modifications to requirements.txt - we just need to tell people in the README to update their `pip` to the latest version before installing from source (virtualenvs, unfortunately, do not install newest pip when creating the environment).

ZenulAbidin:
It works unmodified with updated pip too. Closing this PR.

Issue #217, opened 2022-02-22
The docker image on Docker Hub for version 1.2.7 (latest) has broken auto language detect


MatsBjerin:
To replicate

1. Create this docker-compose file:

```
version: "3"

services:
  libretranslate:
    image: libretranslate/libretranslate:latest
    container_name: libretranslate
    stdin_open: true
    tty: true
    restart: unless-stopped
    ports:
      - 5000:5000
```

2. Start with: "docker-compose up -d"

3. Then run this:
`curl -X POST -H "Content-Type: application/json" -d '{"q": "Ciao!", "source": "auto", "target": "en"}' localhost:5000/translate`

The language auto detect will fail and no translation is made.

cristeigabriel:
I can't try but I recommend that in some 'production' sort of environment you do your posts through an error handle-able module. My guess is that the character is throwing off the language auto detection. Personally I applied the following changes to the LibreTranslate app to 'language.py'. The error should be seen in the host console.

```
RE_BAD_CHARS = regex.compile(r"\p{Cc}|\p{Cs}")

def remove_bad_chars(text):
    return RE_BAD_CHARS.sub("", text)

def detect_languages(text):
    # detect batch processing
    if isinstance(text, list):
        is_batch = True
    else:
        is_batch = False
        text = [text]

    # get the candidates
    candidates = []
    for _t in text:
        t = remove_bad_chars(_t)
```
I have borrowed the solution from someone else who was having the same issue long ago, sadly I didn't store a reference to the post, anyhow. Nonetheless, this solution is still working just fine, after long periods of testing.

cristeigabriel:
If above fixes the issue - I'm willing to make a PR (which'd obviously link the solution original poster).

MatsBjerin:
Thank you Cristei,

The issue however affects all text. It does not matter if the input is just "Ciao" or "Hello". Detection will still fail.
The curly braces are not part of the text. They are just Json wrappers as the POST format is Json and should never reach the actual language engine.

The issue appeared in the latest release. However the previous version with tag "main" turns out to also show the same behaviour.

I am guessing this is because of some change in the language models that are downloaded, as I believe the "main" Docker image itself is unchanged but still has changed its behaviour since just a week ago.

We have worked around the issue by using another language detection engine and thus specify the source and target languages specifically when using libretranslate.

We are of course still hoping the actual problem will be fixed at some point as most users will probably be affected.

Best regards,
  Mats

[...]

Issue #148, opened 2021-10-04
WARNING:polyglot.detect.base:Detector is not able to detect the language reliably.


open-antux:
Today I pulled the docker image (LibreTranslate version: 1.2) on my machine, but when I try to auto detect a language the response is:
```json
[
  {
    "confidence": 0,
    "language": "en"
  }
]
```
If I attach the docker container it gives me this error `WARNING:polyglot.detect.base:Detector is not able to detect the language reliably.`

PJ-Finlay:
Is it ever able to detect the language?

open-antux:
After a couple of container restarting, the language auto detect seems to work. The warning remain but the response is correct now

jasperan:
So, you did nothing but to restart a couple of times? @open-antux 

That's weird that it works for you. I have the same issue, not able to properly detect language.

open-antux:
From what you remember yes, I hadn't made any changes

jasperan:
Then I'll have to reply on langdetect's Python library for language detection instead

BiggA94:
I can reproduce this bug.

This seems to be a Problem with the Docker container itself.
On Startup, the container downloads the language files.
If there isn't a separate volume mount, to save the language files outside of the folder, the container does reload the language files on every startup. (So far so good)

After the language files are loaded, the server seems to require a restart (which is not done inside the docker container).

Workaround would be, to use a volume mount, where the language files are stored and restart the server after the download.

for example, use the following docker compose file:
``` docker-compose
version: "3"

services:
  libretranslate:
    image: libretranslate/libretranslate:latest
    stdin_open: true
    tty: true
    volumes:
      - polyglot_data:/root/.local/share/LibreTranslate/polyglot_data
    ports:
      - 5000:5000
    environment:
      - LT_SUGGESTIONS=true
      - LT_FRONTEND_LANGUAGE_TARGET=de
      - LT_LOAD_ONLY=de,fr,en

volumes:
  polyglot_data:

```

The problem, I have with this is, that there is no "download job" that simply downloads the language and exits afterwards.
(This would be very handy for a init container for Kubernetes).

So my Suggestion for the Docker Image:
* ability to start the image with a new command that only downloads (or updates) the needed files and exits
* update usage of the docker image in the Readme

PR #219, opened 2022-03-04
Fix language detection error

ZenulAbidin:
This closes issue #217 (and #148 as well, which references this exact same issue).

The root cause was load_installed_languages() of argostranslate being called at the top of the file instead of inside a function, this caused the list of installed languages to incorrectly be returned as an empty list.

This commit moves the call inside a function.

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -102,7 +102,7 @@ def create_app(args):
 
     boot(args.load_only)
 
-    from app.language import languages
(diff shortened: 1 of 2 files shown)


pierotofy:
Thanks for the fix @ZenulAbidin ! :pray:

dingedi:
Nice catch !! This will solve a lot of issues

PR #220, opened 2022-03-05
Add the PussTheCat.org instance

TheFrenchGhosty:


dingedi:
Thanks for hosting a LibreTranslate instance !

PR #223, opened 2022-03-09
Fix broken link on README

ewrenge:
I don't know if this is the right way, but I thought broken links are maybe not good


pierotofy:
Indeed they aren't! Thanks!

PR #229, opened 2022-03-25
include language detected, confidence for /translate calls with auto detect

K-Francis-H:
I started using this library recently and thought that this might be a useful feature. Now when using "auto" for the src language on the /translate endpoint, the detected language and confidence value will be included in the JSON result as an object (or array of objects for batch requests) labeled "detectedLanguage" with members: "confidence": <Number> and "language": <string (language code)> . If "auto" is not used the default behavior (only return "translatedText") remains. If "auto" is included the returned object should be backwards compatible with existing consumers of the API, it only adds an additional field that is either one object or an array of objects equal in length to the number of text snippets sent in a batch request.

The reason that I wanted to add this was for a web extension that I am writing. I've noticed that auto detect is slower than specifying a source language and found it useful to get the detected source language(s) from the auto detect feature in the /translate endpoint rather than having to make a separate call to /detect first.

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -446,18 +446,18 @@ def create_app(args):
                 else:
                     # Unable to accurately detect languages for short texts
                     candidate_langs = overall_candidates
-                source_langs.append(candidate_langs[0]["language"])
+                source_langs.append(candidate_langs[0])
 
                 if args.debug:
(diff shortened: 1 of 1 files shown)


pierotofy (approved):
Great addition @K-Francis-H ! :+1:  Thanks.

PR #232, opened 2022-03-26
Upgrade to flask 2

dingedi:
Update flask and other dependencies. It seems to work without any change in the code, but it would probably be necessary to test more.


pierotofy:
Seems to work, and I don't see many breaking changes from 1.x to 2.x in the Flask upgrade notes. :+1: 

Thanks!

PR #235, opened 2022-03-31
Added models include option in docker build

setokesan:
Add the possibility to obtain specific models at docker image build. 
Eq. `docker build -t libretranslate:latest --build-args models=en,jp .`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM python:3.8.12-slim-bullseye
 
 ARG with_models=false
+ARG models=
 
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


pierotofy (approved):
:+1: thanks!! Looks great.

TheFrenchGhosty:
@setokesan This breaks build if no languages are explicitly specified:

```
Updating language models
Found 54 models
Traceback (most recent call last):
  File "install_models.py", line 10, in <module>
    check_and_install_models(force=True, load_only_lang_codes=lang_codes)
  File "/libretranslate/app/init.py", line 34, in check_and_install_models
    raise ValueError(
ValueError: Unavailable language codes: .
```

dingedi:
good catch, it would actually be necessary to set up a CI that verifies each pull request and set up real unit tests. I'll take a look at that.

pierotofy:
Should be fixed with https://github.com/LibreTranslate/LibreTranslate/commit/8cc03572e53a0be3d6ce55c46b76b4c32882f5ed

We don't try to break things on purpose and it can be difficult to catch all possible code paths. Help us improve our CI process if you care about *your* dependent projects not breaking.

PR #237, opened 2022-04-02
Run test for pull requests, add some functional tests

dingedi:
This adds python tests for each PR, this will make it easier to prevent errors in the code.
I haven't yet found how to test the building of docker containers, I don't know too much about github action yet.

I also start to add functional test

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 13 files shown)


dingedi:
I removed python 3.6 because flask limiter 2.2 supports python >= 3.7

dingedi:
I've added a few tests, I'll make more PR as I go along to add more tests, this will avoid introducing regressions unintentionally

dingedi:
Is it good for python 3.6 support removal @pierotofy ? or we downgrade Flask-Limiter

pierotofy:
I think it's OK to remove Python 3.6. :+1: 

Let me know once this is ready and I'll review. Thanks!

dingedi:
this should be good now for python 3.6 :)

dingedi:
@pierotofy I will add more tests later, there is already a small base.
I defined load-only for the tests to en, es to avoid that all the models are downloaded at each test (it was a lot since all this time!)

pierotofy (approved):
Looks great!

Issue #227, opened 2022-03-19
docker-compose build failed. /usr/bin/env: ‘python\r’: No such file or directory


ertopcuburak:
#10 169.0 WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
#10 169.2 /usr/bin/env: ‘python\r’: No such file or directory
------
failed to solve: rpc error: code = Unknown desc = executor failed running [/bin/sh -c if [ "$with_models" = "true" ]; then          pip install -e .;          ./install_models.py;      fi]: exit code: 127

eisenwinter:
this is due to windows lineendings

you can run `git ls-files -z | xargs -0 dos2unix`  on the folder to fix all lineendings to unix linenedings (if youre on windows)

this could be avoided by adding a `.gitattributes` file to the repo containing `*.py text eol=lf ` wich would suggest unix linefeeds to git when pulling or if you dont want to suggest a linefeed you may clean convert the linefeeds when copying files to the container

hope this somewhat helps

PR #246, opened 2022-04-16
Add .gitattributes

DmitrySandalov:
PR for #227.

Just one file has CRLF endings and it's not related to the issue.
```
± % find . -not -type d -exec file "{}" ";" | grep CRLF
./app/static/css/materialize.min.css: ASCII text, with very long lines, with CRLF, CR, LF line terminators
```

`docker-compose build` works fine in Linux, so this must be related to Windows with incorrect or missing `core.autocrlf`.

I agree with @eisenwinter, adding `.gitattributes` should solve the issue.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.py text eol=lf


pierotofy:
Thanks @DmitrySandalov !

PR #250, opened 2022-04-21
Fix badges in the readme

vemonet:
Hi @pierotofy ! Just a small pull request to fix the badges in the readme to remove the old workflows and add the new ones

The workflow to publish the docker images is much nicer now, nice job!


pierotofy:
Thanks @vemonet !

Credit goes to some of our contributors for improving the docker build process :+1:

PR #257, opened 2022-05-03
improve translation formating

dingedi:
_I reverted the previous one because I clicked on merge without doing it on purpose (oops, #255 and #256 ) I wanted to know if you were interested in it before_

This PR improve the formatting of translations.

- If the word to be translated is in uppercase, this change the translation to uppercase (if it was not). Same for lowercase.
- If the word begins with a capital letter or a lower case letter, the same goes for the translation.

Example FR > IT
"**LA MER**" is currently translated into "**il mare**"
"**La viande**" is currently translated into "**carne**"

with this PR:
"**LA MER**" will be translated into "**IL MARE**"
"**La viande**" will be translated into "**Carne**"

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -474,6 +474,21 @@ def create_app(args):
         if text_format not in ["text", "html"]:
             abort(400, description="%s format is not supported" % text_format)
 
+        def improve_translation(source, translation):
+            if source.islower():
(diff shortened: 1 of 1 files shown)


pierotofy:
:+1: thanks!

PR #258, opened 2022-05-06
Update README.md

pingufreak:
File translation via Apache proxy needs "ProxyPreserveHost On", which is "Off" on some distributions (RHEL8 for instance). Else window.location.host will point to the container IP instead of the host.


pierotofy (approved):
:+1: Thanks!!

PR #268, opened 2022-05-30
improve translation of punctuation

dingedi:
this should improve some translations

example: [https://libretranslate.com/?source=fr&target=en&q=bonjour](https://libretranslate.com/?source=fr&target=en&q=bonjour)
"**bonjour**" in French is translated "**Hello.**" in English. with the previous fix the uppercase letters were corrected, with this fix the punctuation is corrected

"**bonjour**" in French will now be translated in "**hello**" in English

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -475,6 +475,21 @@ def create_app(args):
             abort(400, description="%s format is not supported" % text_format)
 
         def improve_translation(source, translation):
+            source = source.strip()
+
+            source_last_char = source[len(source) - 1]
+            translation_last_char = translation[len(translation) - 1]
+
+            punctuation_chars = ['!', '?', '.', ',', ';']
+            if source_last_char in punctuation_chars:
+                if translation_last_char != source_last_char:
+                    if translation_last_char in punctuation_chars:
+                        translation = translation[:-1]
+
+                    translation += source_last_char
+            elif translation_last_char in punctuation_chars:
+                translation = translation[:-1]
+
             if source.islower():
                 return translation.lower()
 


pierotofy:
Nice! I wonder if long-term there's a way to encode this logic into the language model, but for now this should help with languages like French.

dingedi:
I don't know if we can add logic in the models (ping @PJ-Finlay ) I think we just need to improve the models with more and better data sets

this seems a problem with many languages:

- "hello" -> "Hola." (en->es)
- "hello" -> "Hallo." (en->nl)
- "hello" -> "Olá." (en->pt)
- "hello" -> "Pronto?" (en->it)
- "hello" -> "Hallo" (en->de) no punctuation problem but a capital letter is added

PJ-Finlay:
The models are mapping Unicode in the source language to Unicode in the target language so it could be improved to better handle capitalization. This could be done by training on data that handles capitalization with the desired behavior.

We could either add new data to the [LibreTranslate Community Datasets](https://github.com/LibreTranslate/CommunityDS) or filter existing data like this commit is doing. I've moved away from doing automated data modifications in the Argos Translate training scripts because it's slow. If we wrote a set of filters/rules we could process existing data to create a new dataset then train on that data.

SethFalco:
Hmm, is there documentation anywhere with what sources Argos Translate / LibreTranslate is using to create the datasets the model(s) are trained on?

I'm curious what data is there and how large it is.

I don't know if it'll be much, but we could consider looking into how to export all approved strings on [the public/hosted instance of Weblate](https://hosted.weblate.org/) and include that in a dataset? If that'd be useful, I'd happily look into how to dump this?

Then the script and dump can be saved somewhere publicly accessible. (And separated by license if that's a concern.)

PJ-Finlay:
The data mostly comes from [Opus](https://opus.nlpl.eu/) currently but if we scrape other sources or build new datasets we could use them for training future models.

PR #272, opened 2022-06-08
Update README.md

pooyaEst:
`docker build [--build-arg with_models=true] -t libretranslate .` causes `zsh: bad pattern: [--build-arg` error

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,10 +145,10 @@ Then open a web browser to http://localhost:5000
 ### Build with Docker
 
 ```bash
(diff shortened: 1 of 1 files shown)


dingedi:
The provided command are for Bash, if you use zsh you need to adapt it, i think its a bad Idea to replace the Bash commands for zsh commands in the doc.

pierotofy:
Agreed we shouldn't remove the bash command; perhaps this PR can be improved to inform zsh users of the different syntax?

pierotofy:
Closing, feel free to re-open a new one :pray:

PR #276, opened 2022-06-21
fix: fix race condition in settings and langs

SethFalco:
I ran into a problem on the website where the `source` and `target` in the query parameters would be overridden by the default settings sometimes.

It's because of a race condition with both the `/frontend/settings` and `/languages` requests modifying the same variables when they finish.

Sometimes `/languages` will finish first, which has the handling for the query parameters. Then when the `/frontend/settings` request finished, it'd remove the user-configured source and target.

This resolves the issue by still doing both requests asynchronously, but we stall the handling of the `/languages` endpoint until the `/frontend/settings` request has been handled.

---

On the side this makes some other changes:
* Formatting
* Replace XMLHttpRequest with `fetch`. (All modern browsers have supported it since 2015/2016.
* Replace the basically duplicate code with a single function with parameters.
* Moves `self.langs.push({ name: 'Auto Detect (Experimental)', code: 'auto' })` to _below_ the check for the number of languages. Before there could never be 0 languages because the `auto` language was pushed into the array before the check.
* Removes the check for `isSourceAuto`, because an extra language is added which already has the code `auto`, this was already handled.

### Related
* Closed https://github.com/LibreTranslate/LibreTranslate/pull/279

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -39,15 +39,18 @@ document.addEventListener('DOMContentLoaded', function(){
             filesTranslation: true,
             frontendTimeout: 500
         },
-        mounted: function(){
-            var self = this;
-            var requestSettings = new XMLHttpRequest();
-            requestSettings.open('GET', BaseUrl + '/frontend/settings', true);
+        mounted: async function() {
+            const self = this;
 
-            requestSettings.onload = function() {
-                if (this.status >= 200 && this.status < 400) {
-                    // Success!
-                    self.settings = JSON.parse(this.response);
+            const settingPromise = fetch(BaseUrl + "/frontend/settings");
+            const langsPromise = fetch(BaseUrl + "/languages");
+
+            try {
+                const settingsResponse = await settingPromise;
+
+                if (settingsResponse.status >= 200 && settingsResponse.status < 400) {
+                    self.settings = await settingsResponse.json();
+
                     self.sourceLang = self.settings.language.source.code;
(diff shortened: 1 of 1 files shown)


pierotofy:
Hey @SethFalco :hand: thanks for the fix! The purpose of using `XMLHttpRequest` was to maximize browser compatibility; some older mobile browsers and some older browsers (less important) do not support it. Same for the use of `async`, so I cannot merge this one as-is.

Is it possible to fix the race condition without all of the refactoring?

SethFalco:
Yeah, it'll definitely be possible, though I'll have to peek at what's the nicest way to do it with XMLHttpRequest since I don't use it much. 🤔 

Are you sure about wanting to avoid `fetch` and `async` though? I was under the impression they were pretty well established for a while now?

Fetch API:
* https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility
* https://caniuse.com/fetch

On MDN and caniuse it shows very broad support:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#browser_compatibility
* https://caniuse.com/async-functions

Note that Baidu Browser is discontinued and deprecated in 2019. ([Reference](https://en.wikipedia.org/wiki/Baidu_Browser))

---

If you're confident, I'll swap out but I do think it'd be worthwhile to move to a more modern API to keep things tidy. ^-^'

PR #279, opened 2022-06-22
fix: fix race condition in settings and langs

SethFalco:
This is a duplicate PR of https://github.com/LibreTranslate/LibreTranslate/pull/276 with a different approach.

This one doesn't use `fetch` or `async`.

Basically we do both requests at the same time, then we check if `langsRequest.response` is set. If it is, that means the request had finished so we just pass that, but if not then we add the `onload` handler.

We can't just depend on `onload` here because if it's outside (like before) then it might execute first which causes the issue.

If we add `onload` in the handler for `settingsRequest`, it won't trigger if the request is finished before the `onload` handler is added.

---

I ran into a problem on the website where the `source` and `target` in the query parameters would be overridden by the default settings sometimes.

It's because of a race condition with both the `/frontend/settings` and `/languages` requests modifying the same variables when they finish.

Sometimes `/languages` will finish first, which has the handling for the query parameters. Then when the `/frontend/settings` request finished, it'd remove the user-configured source and target.

This resolves the issue by still doing both requests asynchronously, but we stall the handling of the `/languages` endpoint until the `/frontend/settings` request has been handled.

---

On the side this makes some other changes:
* Formatting
* Replace the basically duplicate code with a single function with parameters.
* Moves `self.langs.push({ name: 'Auto Detect (Experimental)', code: 'auto' })` to _below_ the check for the number of languages. Before there could never be 0 languages because the `auto` language was pushed into the array before the check.
* Removes the check for `isSourceAuto`, because an extra language is added which already has the code `auto`, this was already handled.

---

Closes https://github.com/LibreTranslate/LibreTranslate/pull/276

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -39,14 +39,17 @@ document.addEventListener('DOMContentLoaded', function(){
             filesTranslation: true,
             frontendTimeout: 500
         },
-        mounted: function(){
-            var self = this;
-            var requestSettings = new XMLHttpRequest();
-            requestSettings.open('GET', BaseUrl + '/frontend/settings', true);
+        mounted: function() {
+            const self = this;
 
-            requestSettings.onload = function() {
+            const settingsRequest = new XMLHttpRequest();
+            settingsRequest.open("GET", BaseUrl + "/frontend/settings", true);
+
+            const langsRequest = new XMLHttpRequest();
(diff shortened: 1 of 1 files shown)


pierotofy (approved):
Looks great @SethFalco :pray: thanks so much for the fixes!

PR #283, opened 2022-07-03
Expand CUDA Support to more GPUs

jonwiggins:
Hi,

After my last PR I noticed that the container would fail to translate on some GPU types.
Specifically, it would fail with:
```
{GPU} with CUDA capability {sm_**} is not compatible with the current PyTorch installation. 
The current PyTorch install supports CUDA capabilities sm_37 sm_50 sm_60 sm_70.
```

This is because one of the other package requirements installs the default version of PyTorch, which is only setup for some GPUs.
This change explicitly installs a CUDA PyTorch version which allows for more GPUs to be supported.

See also: https://discuss.pytorch.org/t/need-help-trouble-with-cuda-capability-sm-86/120235

While I haven't tried all of the GPUs, I was able to get this running on an A10 GPU.

A list of GPUs and their compute capability versions can be found in the Nvidia documentation: https://developer.nvidia.com/cuda-gpus

Let me know what you think, I'm happy to make any needed changes.

Thanks,
Jon

### DIFF
--- a/docker/Dockerfile.cuda
+++ b/docker/Dockerfile.cuda
@@ -18,6 +18,8 @@ COPY . .
 
 RUN ln -s /usr/bin/python3 /usr/bin/python
 
+RUN pip3 install torch==1.12.0+cu116 -f https://download.pytorch.org/whl/torch_stable.html
+
 RUN if [ "$with_models" = "true" ]; then  \
     # install only the dependencies first
     pip3 install -e .;  \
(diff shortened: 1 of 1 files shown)


pierotofy (approved):
Thanks @jonwiggins ! Looks great.

PJ-Finlay:
I think this is a similar issue, if the Pytorch version doesn't match the CUDA version you can get issues. 

- https://forum.opennmt.net/t/cuda-error-no-kernel-image-is-available-for-execution-on-the-device/4644/3

PJ-Finlay:
We could also consider disabling CUDA for Stanza to avoid this type of issue. I think Stanza is the only dependency that needs Pytorch.

- https://github.com/argosopentech/argos-translate/issues/274#issuecomment-1151672350

PR #285, opened 2022-07-15
Add --api-keys-db-path argument

Minosity-VR:
# Issue
The API keys database is currently directly under `/app`. This make it hard to manage it with docker volumes without targeting the whole `/app`: 

Targeting direclty the `api_keys.db` file would be bound to fail:
> The bind-mount is based on inode on Linux. Most of the time, when we change/edit a file, the editor will create a new inode and replace the old file with the new file. In this case, the bind-mount is still on the old file and the container do not see the change. ([From a docker forum post](https://forums.docker.com/t/sharing-a-single-file-via-a-volume-mount-does-not-work-as-expected/57088/3))

# What does this PR ?
Create a optional argument `--api-keys-db-path` to choose a path inside the container to store the database.

The default value is the previous value `api_keys.db` to ensure compatibility with currently deployed instances.

# How was it tested ?
I tried to run the docker without the argument, with absolute path (`/app/db/api_keys.db` and `/root/.local/db`), a relative path (`./db/api_keys_different_name.db` and `db/api_keys.db`), and relative path without folder (`api_keys.db`, `./api_keys.db` and `different_name.db`) with the command
```bash
docker run -it -p 5000:5000 -v libretranslate_models:/root/.local libretranslate --api-keys --req-limit 2 --api-keys-db-path <path>
```

Each time I checked that the route `/translate` was working with and without an API key.

It is my first time contributing to an open project, I'm sorry if I missed something (such as a PR template or some guidelines), feel free to request changes is something is bothering you :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,8 @@ docker-compose up -d --build
 
 > Feel free to change the [`docker-compose.yml`](https://github.com/LibreTranslate/LibreTranslate/blob/main/docker-compose.yml) file to adapt it to your deployment needs, or use an extra `docker-compose.prod.yml` file for your deployment configuration.
 
+> The models are stored inside the container under `/root/.local/share` and `/root/.local/cache`. Feel free to use volumes if you do not want to redownload the models when the container is destroyed. Be aware that this will prevent the models from being updated!
+
 ### CUDA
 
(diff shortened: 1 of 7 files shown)


pierotofy (approved):
Thanks for the PR @Minosity-VR :+1: This looks fantastic.

PR #293, opened 2022-08-06
brightened the Realm of Darkness

ecxod:
some CSS adjustments to make "Tanslate-Files" look like "Translate-Text in dark mode


pierotofy:
:pray: thanks @ecxod !

PR #294, opened 2022-08-06
Preloading stuff

ecxod:
here I propose to preload the static stuff so the loading gets faster. i also placed the stylesheets at the end of the header because they get preloaded after the first loading of the site.
after line 51 there is just vertically moving code to improve readable.


pierotofy:
Thanks @ecxod :+1:

Issue #303, opened 2022-08-24
Switch Languages Button not visible in Dark Mode


youngtrashbag:
This is just a small display problem:
![grafik](https://user-images.githubusercontent.com/20374823/186348883-298f3087-4ebd-41c7-93bb-5a28db3e67ef.png)

I fixed it by adding `color: white;` to the element:
![grafik](https://user-images.githubusercontent.com/20374823/186349297-98ce97f6-bf00-44ef-b23a-a8fabc88cd95.png)

I'm not very experienced with CSS, so I don't know if I can open a pull request for this, but I will research it later (detection of theme, etc).

youngtrashbag:
Merged in #304

PR #304, opened 2022-08-24
Make switch-language-btn white in Dark Theme

youngtrashbag:
MR to issue #303 

Theming was already handled, so "fix" was very easy.


youngtrashbag:
I just noticed that by removing the `color` attribute, the color defaulted to `#039be5`, which is the default for all `<a>` tags, so I'm not sure which color would fit better.

pierotofy:
Thanks! 👍

PR #312, opened 2022-09-13
Non-root User for Dockerfile and Multi-Stage Build

eugene-davis:
Updates the Dockerfile to improve security and image size.
In particular, this allows the image to be run under the "restricted" pod security standard in Kubernetes.

* Creates and installs under a new user, `libretranslate` to avoid requiring root
* Splits the Dockerfile into a multistage to to avoid keeping dev tools in the final image
* Updates the base image to 3.8.14

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,5 @@
 FROM python:3.8.12-slim-bullseye as builder
 
-ARG with_models=false
-ARG models=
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks @eugene-davis ! :pray: looks great.

PR #317, opened 2022-09-23
move improve_translation in language.py and use it in transliteration

### LINKED ISSUE #316: Bad translations due to transliteration
I noticed that some translations are different between [argos-translate](https://github.com/argosopentech/argos-translate) and libretranslate, after some research it comes from the transliteration with [polyglot](https://github.com/aboSamoor/polyglot)

With argos-translate directly:

```
$ argos-translate --from-lang it --to-lang fr 'DULUX L'
DULUX L
```

With libretranslate
image

Printing the source text and that transliterated, we see that it has been modified (uppercase to lowercase):
```
source: DULUX L
transliterated: dulux l
```

of course argos-translate should not translate into musical note even if the word is in lowercase but that's another problem with models

dingedi:
This shoud close #316 by restoring the uppercase lowercase formatting of the word before transliteration with polyglot.
but polyglot has not had a release since 2016 it may be necessary to change to a more recent updated library.

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -15,7 +15,7 @@ from translatehtml import translate_html
 from werkzeug.utils import secure_filename
 
 from app import flood, remove_translated_files, security
-from app.language import detect_languages, transliterate
+from app.language import detect_languages, transliterate, improve_translation_formatting
 
 from .api_keys import Database, RemoteDatabase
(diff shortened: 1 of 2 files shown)


pierotofy:
Yeah long term a different approach for language detection will be needed. Thanks for the refactoring, makes the code more modular.

Issue #314, opened 2022-09-19
Auto Detect should print the detected language


89z:
This URL:

https://libretranslate.com?source=auto&target=en&q=暂时保存

the result just says:

> Auto Detect (Experimental)

Other sites:

https://deepl.com

will print the detected language:

> Chinese (detected)

dingedi:
yes it would be a good improvement

89z:
How long does it take to go live?

PR #323, opened 2022-10-01
Show detected Language (#314)

AnTheMaker:
Hi there!
This is a minimal implementation of #314.
It should print the detected language as well as the confidence when using "Auto Detect" as the source language.
Feel free to see this as a rough starting-point - please add your own suggestions and changes to this PR!
~ An

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -25,6 +25,8 @@ document.addEventListener('DOMContentLoaded', function(){
             translatedText: "",
             output: "",
             charactersLimit: -1,
+            
+            detectedLangText: "",
 
             copyTextLabel: "Copy text",
 
@@ -189,6 +191,8 @@ document.addEventListener('DOMContentLoaded', function(){
                 if (this.timeout) clearTimeout(this.timeout);
                 this.timeout = null;
 
+                this.detectedLangText = "";
+                
                 if (this.inputText === ""){
                     this.translatedText = "";
                     this.output = "";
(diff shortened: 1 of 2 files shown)


pierotofy (approved):
Thanks @AnTheMaker ! Looks great.

app/static/js/app.js
```diff
@@ -223,6 +227,9 @@ document.addEventListener('DOMContentLoaded', function(){
                                 self.translatedText = res.translatedText;
                                 self.loadingTranslation = false;
                                 self.output = JSON.stringify(res, null, 4);
+                                if(self.sourceLang == "auto" && res.detectedLanguage.length > 0){
```
dingedi, line 230: there is a problem here, detectedLanguage is an object we cannot use .length because it returns undefined on an object
so the condition never triggers

app/templates/index.html
```diff
@@ -142,6 +142,7 @@ <h3 class="header center">Translation API</h3>
 						<div class="row mb-0">
 							<div class="col s6 language-select">
 								<span>Translate from</span>
+                <span><i>[[ detectedLangText ]]</i></span>
```
dingedi, line 145: I think it should rather display the language and not the ISO code

app/static/js/app.js
dingedi:
```js
if(self.sourceLang == "auto" && res.detectedLanguage !== undefined){
```

this solve the problem

pierotofy: Ah, good catch @dingedi . I should have reviewed more thoroughly. Can you make a commit into main?

dingedi: yes done in https://github.com/LibreTranslate/LibreTranslate/commit/5d8e513d45e4764189b3b7516e5b9f29e6a6f38e

AnTheMaker: Thanks for noticing & fixing @dingedi!

app/templates/index.html
dingedi: done in https://github.com/LibreTranslate/LibreTranslate/pull/324

AnTheMaker: Perfect! Was working on this improvement too at the moment! Thank you! :)

PR #329, opened 2022-10-05
Add shell binding to README

Hayao0819:
Add bindings for Shell Scripts

Wrote functions for LibreTranslate that can be used in the shell. It is very small, but hopefully it will be useful for someone.

The license is WTFPL, so it is freely available to everyone.


pierotofy:
Very cool scripts, thank you for sharing @Hayao0819 !

PR #334, opened 2022-10-30
Update main.css

ecxod:
do not eat the first Letter on Firefox

image


pierotofy:
Hey @ecxod :wave: thanks for looking into this; what browser (which Firefox version) and OS did you test this on? I ask because I currently can't see the clipping shown in the screenshot above on Firefox on Windows.

image

SethFalco:
I guess this issue myself, actually.

image

OS: Debian 11
Browser: Firefox 108.0a1 (2022-10-21) (64-bit)

ecxod:
@pierotofy @SethFalco Please open a new [project](https://github.com/LibreTranslate/LibreTranslate/projects) for mobile. There are a lot of issues and here is the wrong place to talk about them.
We must make a lot of small css for different devices and then compile them into one css. lets hit the horses.

pierotofy:
How about just opening a dedicated issue and making a list of the outstanding things to do?

PR #362, opened 2022-12-11
ARM32 build

pierotofy:
As the title says.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Build and push Image
         uses: docker/build-push-action@v2
         with:
-          platforms: linux/amd64,linux/arm64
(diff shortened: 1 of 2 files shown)


lordyavin:
Hey @pierotofy I tried to build the image on a 32bit ARMv7 environment (Raspberry Pi 4 with 32bit RaspberryPi OS) first issue I had was missing development packages:

* libxml2-dev
* libxslt-dev 
* python-dev

I added `RUN apt-get install -y libxml2-dev libxslt-dev python-dev` 

The second issue I ran into and what seams to be a blocker is that ctranslate is not available for 32bit ARM architecture:

https://pypi.org/project/ctranslate2/2.24.0/
> The project supports x86-64 and AArch64/ARM64 processors

Is there any change to workaround this? Unfortunately I've overseen the point that the PI imager uses 32bit OS as default 😠

PR #366, opened 2022-12-20
Run helper script, misc improvements

pierotofy:
- [x] Adds `run.sh` script for easier docker setup that doesn't need to re-download language models at startup by default.
 - [x] Adds `run.bat` script (Windows)
 - [x] Adds `--update-models` command line argument, which allows language model updates before boot.


Saijin-Naib:
I was just trying to figure out how to adapt your run.sh script from WebODM to LibreTranslate since I was trying to stand up my own instance.

This is very, very nice for ease of use!

PR #368, opened 2022-12-26
Prometheus Support

pierotofy:
This PR adds support for making LibreTranslate a Prometheus exporter.

Prometheus can be used to collect usage metrics from LibreTranslate (and a ton of other apps). Data from prometheus can then be displayed using Grafana and many other tools.

By default this is disabled. Enable it with `--metrics` and secure it (optionally) with `--metrics-auth-token`.


pierotofy:
Example dashboard:

image

PR #372, opened 2022-12-29
app->libretranslate; mv tests/ inside libretranslate/

### LINKED ISSUE #371: `pip install` makes `app` module (not `libretranslate`)
After installing `libretranslate` using `python3 -m pip install libretranslate`, the package is successfully installed but the structure is not as expected, and actually uses bad practices. For example, the main `libretranslate` executable imports the app using `from app.main import main`, but after installing `libretranslate` the expected behavior is to do something like `from libretranslate.app import main`. Such a generic package name like `app` is likely to cause conflicts with many people's code.

I suggest refactoring so that the package is all contained inside a `libretranslate` module.

reynoldsnlp:
fixes #371


reynoldsnlp:
I moved `tests/` inside the `libretranslate` module, so they would not be installed as their own module. Another approach would be to leave the `tests/` dir in the root dir, but to delete the `tests/__init__.py` files.

pierotofy (approved):
Thanks for the improved directory/module structure @reynoldsnlp, this looks great!

PR #374, opened 2022-12-31
Allow `url-prefix`

### LINKED ISSUE #373: Allow `url-prefix`
Allow adding prefix to the URL, such as `http://example.com:5000/url-prefix`.

reynoldsnlp:
Fixes #373


pierotofy:
Thanks @reynoldsnlp, I know the ability to add a prefix has been requested a few times so this is super-useful.

A few things I found that could be addressed before I merge it:

 - [ ] The link to the "API Docs" (/docs) breaks when accessing the page from a prefix.
 - [ ] Accessing the UI from the prefix, typing in a translation, then refreshing the page causes the app to hang with: 
```
Uncaught SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
    at JSON.parse (<anonymous>)
    at settingsRequest.onload (app.js?v=1.3.7:56:42)
settingsRequest.onload @ app.js?v=1.3.7:56
load (async)
mounted @ app.js?v=1.3.7:54
Be @ vue@2.js:6
Qt @ vue@2.js:6
(anonymous) @ vue@2.js:6
Cn.$mount @ vue@2.js:6
Cn.$mount @ vue@2.js:6
t._init @ vue@2.js:6
Cn @ vue@2.js:6
(anonymous) @ app.js?v=1.3.7:10
```

reynoldsnlp:
@pierotofy I think I've fixed both of those issues. I'm not doing much testing myself, so see if you can find any other bugs before merging.

pierotofy:
Looks great, thanks!!

simao-silva:
How should the syntax be when using WSGI and Gunicorn? Currently, I am starting this way
 ```
 SCRIPT_NAME=/translate ./venv/bin/gunicorn --bind 0.0.0.0:5000 --log-level 'DEBUG' 'wsgi:app'
 ```
 but this does not change the Swagger URL, for example.

pierotofy:
Could we move this conversation over to the forum at https://community.libretranslate.com? :pray: The forum is the right place to ask questions (we try to keep the GitHub issue tracker for feature requests and bugs only). Thank you! :+1:

PR #382, opened 2023-01-06
Use WAI-ARIA attributes to make the web interface more accessible

jmdaweb:
The following changes will improve experience for screen reader users:

* Alt attribute content has been removed on application logo. Now, the alt attribute is empty, marking the image as decorative. The link is descriptive enough with its current text content.
* Several items, such as some buttons and links, have been labelled using aria-label.
* In other input fields, such as source and target language combo boxes, an id has been added to their above span elements. This id is used in a new aria-labelledby attribute.
* Aria-hidden attribute has been added to some items.

Note that there are more clean and native techniques to improve accessibility without ARIA, but I'm blind and I'm sure ARIA won't have a visual impact that I won't be able to test.


pierotofy:
Thanks @jmdaweb :pray: this looks great!

PR #394, opened 2023-01-15
added script to allow healthcheck call in docker-compose

jonathan-chin:
when launching a new container, it was always difficult to tell when the container had finished downloading all the language models. I created a simple python script to make a REST call and try to translate English to English (assuming that the user, at the very least, has the English language model installed).

calling this allows us to now check if the server is healthy / unhealthy. I've also updated `docker-compose.yml` and `README.md` with an example of how to use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,15 @@ Run this version with:
 docker-compose -f docker-compose.cuda.yml up -d --build
 ```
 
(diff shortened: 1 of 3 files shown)


pierotofy:
Hey @jonathan-chin :wave: thanks !

Issue #395, opened 2023-01-17
General issue about bad language detection (and how to possibly fix it)


schrmh:
(Does address issues like https://github.com/LibreTranslate/LibreTranslate/issues/247 as well)

So, this night I tried out LibreTranslate again cause a friend said I should take a look. I found out it used polyglot (was corrected later that it now uses CLD 2 directly instead of polyglot which uses CLD2 via another library) which just performs horrible when it comes to the languages I care about (German, Japanese and English).
I then noticed that LibreTranslate used to use langdetect which was way more accurate when it came to my personal tests thus I commented on a merge request which also contains some findings so far:
https://github.com/LibreTranslate/LibreTranslate/pull/60#issuecomment-1384644446 contains my findings so far (read following comments as well there)

After this, I found out a bit more (e.g. for German and English, fasttext seems to be actually better according to graphs in https://github.com/pemistahl/lingua-py. 
And when it comes to quotes then lingua might fail https://github.com/pemistahl/lingua-py/issues/112 when it would return better results than langdetect if it was asked for each part individually) and thus I'm opening a issue to make it easier to keep an eye on that and cases which should be considered (and to attract potential contributers).

So things to do IMO:
- Replace CDL 2 with lingua-py for a quick boost in accuracy (it seems it is only worth to use CDL 2 if Mangolian is one of the languages you want to detect and the others are languages in which CDL 2 is as good as the other libraries)
- Implement fasttext as another option and bring back langdetect as well
- Give the user the choice which to use
- load-only should then really only care about selected languages if possible (I have the feeling it doesn't, looking at accuracy)
- Try to choose the library based on the load-only languages
- Try to choose the library depending on what is input (e.g. single word, sentence, sentences with words in other languages quoted like `Er sagte "わたし"`)
- Option to be deterministic if possible (e.g. langdetect provides an option like this so that it doesn't detect something as language 1 in the first try and the very same thing as language 2 in another try)

Detection might be tested by looping over words and sentences from lists from wiktionary (it provides such lists in machine-readable formats as downloads: https://en.wiktionary.org/wiki/Help:FAQ#Downloading_Wiktionary) and then checking out the cases it did not detect as expected (e.g. is there a different language with the same word or was it really just bad detection?).

More complex but since e.g. lingua supports multi language detection in mixed texts it might be worth it to consider to implement such things as well...:
- Option to enable/disable multi language detection in input
[...]

PR #396, opened 2023-01-17
Improve Language detection

### LINKED ISSUE #247: French word not recognised
Hi,
I have used the translate.argosopentech.com API to translate french into english but got the same french sentence returned.

The sentence was: "Tout philosophe a deux philosophies : la sienne et celle de Spinoza."

As soon as I cut "la sienne et celle de spinoza" off the text, I get a successful french to english translation. This isn't the first I get this bug. Maybe due to a small french dataset on which libretranslate has been trained on...

Joshix-1:
Try to fix some of the issues raised in #395 and fix #247

- Replace CDL 2 with lingua-py for "a quick boost in accuracy"
- ingore languages that are not loaded for language detection

I didn't remove the old detection class, so maybe later multiple methods can be used.

With these changes `Tout philosophe a deux philosophies : la sienne et celle de Spinoza.` can be translated when is language is auto.
Even single german words like "Pfannkuchen" get recognised.

### DIFF
--- a/libretranslate/app.py
+++ b/libretranslate/app.py
@@ -24,6 +24,7 @@ from libretranslate.locales import (_, _lazy, get_available_locales, get_availab
         gettext_html, lazy_swag, get_alternate_locale_links)
 
 from .api_keys import Database, RemoteDatabase
+from .detect import Detector
 from .suggestions import Database as SuggestionsDatabase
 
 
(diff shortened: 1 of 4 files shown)


pierotofy:
Hey thanks @Joshix-1 !

It looks like this seems to struggle with single words; percentages are also incorrect.

image

If the intent is to remove cld2, let's just remove it; no need to keep it around. One can always browse git to find the previous code.

Joshix-1:
I fixed the bug with the confidence percentage. 
> It looks like this seems to struggle with single words
Seems to be the case for some. For others it detects them more reliable. Language detections seems to be a difficult task

Joshix-1:
know it doesn't detect the french :(

Joshix-1:
lingua-py seemed to me the better and easier option.
- According to the graphs in the lingua-py README it is better than fasttext.
- on https://fasttext.cc there is no easy usage tutorial

I'm curious how lingua-py would perform with your test-data, so we could get an independent insight to whether it is better or worse than fasttext.

Jourdelune:
With my experience, you must combine dictionary based language detection and different library with ngram.

I have made this library to detect discord message with a good accuracy:

https://github.com/Interaction-Bot/LanguageDetection

Jourdelune:
Thank you for the advice! I didn't do metrics but I have used this paper to make the detector: https://aclanthology.org/W15-1703.pdf.

I have also my own dataset of discord message and I have reached an accuracy of 89% of good detection, and 4% of wrong detection if I use the reliability. I have do a medium post about that: https://medium.com/@jourdelune863/improving-language-detection-for-chat-discord-b148dfacc2e5.

I have trained a custom model of fastText on my type of message to have this reliability. But for you case I think just combine two detector can be enough.

Jourdelune:
yes it's true but detection by dictionnary and add up the results of different detector to detect short text can help you if the accuracy is a problem :p

pierotofy:
Closing in favor of #526 :pray:

PR #398, opened 2023-01-31
Switches (back) to docker/Dockerfile as default for docker-compose

lordyavin:
I assume to use the cuda.Dockerfile as default (via docker-compose.yml) was a mistake.


dingedi:
Thanks nice catch

Issue #359, opened 2022-12-10
Docker image size has doubled after v1.3.0


thomas-mc-work:
Hello team,

I've just spotted that the image size of the Docker image has doubled after `v.1.3.0`:

- `v.1.3.0`: 872 MB
- `v.1.3.1`: 1.8 GB

Are you aware of that resp. was that on purpose?

pierotofy:
Mm, I didn't check (docker builds are automated), but I'm also not sure what caused the increase in size. We'd love help to find out and improve the size.

pierotofy:
There we go:

```
463M    site-packages/nvidia/cublas
463M    site-packages/nvidia/cublas/lib
860M    site-packages/nvidia/cudnn/lib
861M    site-packages/nvidia/cudnn
1.4G    site-packages/nvidia
1.8G    site-packages/torch
1.8G    site-packages/torch/lib
```

ngosang:
I think we should split the Docker image in 2 images: with and without cuda support.
I can open a PR but I will like to hear your opinion first.

dingedi:
Good idea i think. Not everyone need cuda

pierotofy:
I'm OK with that.

Joly0:
Would it be possible to load some of that when the container starts and save it somewhere in a volume? That way those big files only get downloaded once and can be reused every time without stuffing them in the image itself and bloating it that way

thomas-mc-work:
Hello @Joly0,

as a user I always prefer a transparent solution like the split images like suggested above. There you can see beforehand what you can expect. And you know the data transfer is done after you've downloaded the image. For me, it's rather unusual if a download starts on the first startup. Also, this wouldn't heal the problem of having too much data that is being transferred. It's just "hidden" in a delayed transfer. I hope this helps.

Joly0:
> Hello @Joly0,
> 
> as a user I always prefer a transparent solution like the split images like suggested above. There you can see beforehand what you can expect. And you know the data transfer is done after you've downloaded the image. For me, it's rather unusual if a download starts on the first startup. Also, this wouldn't heal the problem of having too much data that is being transferred. It's just "hidden" in a delayed transfer. I hope this helps.

You might have misunderstood what i meant or maybe i wasnt clear enough. 
[...]

PR #425, opened 2023-04-02
Reduce docker image size

TheTakylo:
Fix https://github.com/LibreTranslate/LibreTranslate/issues/359

this PR reduce the docker image size by installing torch with cpu only support https://github.com/LibreTranslate/LibreTranslate/issues/359#issuecomment-1407736532

this also publish a specific image for cuda using the [cuda.Dockerfile](https://github.com/LibreTranslate/LibreTranslate/blob/main/docker/cuda.Dockerfile)
there will be the suffix "-cuda" after the tag name for images with cuda

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -43,3 +43,15 @@ jobs:
           cache-to: type=gha,mode=max
         env:
           TAG: ${{ startsWith(github.ref, 'refs/tags/') && steps.get-variables.outputs.version || 'latest' }}
+
+      - name: Build and push Cuda Image
+        uses: docker/build-push-action@v2
+        with:
+          file: ./docker/cuda.Dockerfile
+          platforms: linux/amd64,linux/arm64
+          tags: ${{ steps.get-variables.outputs.gh-username-lower }}/libretranslate:${{ env.TAG }}-cuda
+          push: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+        env:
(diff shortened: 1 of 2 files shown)


pierotofy:
Thanks @TheTakylo !

gaby:
@pierotofy Thank you for this! 🙌

pierotofy:
Thank @TheTakylo :) I didn't do much!

ngosang:
https://github.com/LibreTranslate/LibreTranslate/actions/runs/4590106961/jobs/8105441967

gaby:
May need to update the tag:

```#22 8.061 ERROR: Could not find a version that satisfies the requirement torch==1.12.0+cu116 (from versions: 1.7.1, 1.8.0, 1.8.1, 1.9.0, 1.10.0, 1.10.1, 1.10.2, 1.11.0, 1.12.0, 1.12.1, 1.13.0, 1.13.1, 2.0.0)```

TheTakylo:
it works well on my pc, I did a test by removing arm64 for the docker cuda version in my github action and the problem seems to come from here.

ngosang:
pyTorch does not provide ARM64 package for CPU => https://download.pytorch.org/whl/torch/
For now we should publish this:

CPU / latest:
* x64

CUDA / latest-cuda
* x64
* ARM64

ngosang:
@TheTakylo If we change the cuda base image we can have both architectures with cuda too.
image

TheTakylo:
the current cuda image support also both architectures, the problem come from torch i think


image

TheTakylo:
it work now

https://github.com/LibreTranslate/LibreTranslate/actions/runs/4597995177/jobs/8121422270

PR #427, opened 2023-04-03
Fix docker cuda image build

TheTakylo:
- Upgrade Babel from `2.11.0` to `2.12.1`
- Disable installation of all languages models by default
- Remove arm64 build for docker cuda image

https://github.com/LibreTranslate/LibreTranslate/pull/425


pierotofy:
Thanks @TheTakylo !

PR #428, opened 2023-04-03
Fix bad help message in manage.py

mnzaki:


pierotofy (approved):
Nice catch! Thanks @mnzaki

PR #434, opened 2023-04-27
Fix compatibility with flask-babel v3

misuzu:
Babel.locale_selector was [removed](https://github.com/python-babel/flask-babel/blob/69d3340cd0ff52f3e23a47518285a7e6d8f8c640/CHANGELOG) in 3.0.0.

### DIFF
--- a/libretranslate/app.py
+++ b/libretranslate/app.py
@@ -1079,13 +1079,14 @@ def create_app(args):
         return jsonify(lazy_swag(swag))
 
     app.config["BABEL_TRANSLATION_DIRECTORIES"] = 'locales'
-    babel = Babel(app)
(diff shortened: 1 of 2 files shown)


pierotofy:
Thanks @misuzu :+1: is this compatible with flask-babel v2? (That's the version that's currently being used and I don't see changes in the requirements.txt).

misuzu:
> Thanks @misuzu +1 is this compatible with flask-babel v2? (That's the version that's currently being used and I don't see changes in the requirements.txt).

It isn't unfortunately, I've updated Flask-Babel to 3.1.0 in requirements.txt

pierotofy:
Looks good, thanks!

PR #440, opened 2023-05-22
Speed up Docker image build and make Docker image smaller

PeterDaveHello:
- `docker/Dockerfile` uses multi-stage build, so the image size is not affected.
- `docker/cuda.Dockerfile` image size change is huge as below, almost 2GB saved:

```
REPOSITORY          TAG                  IMAGE ID       CREATED         SIZE
after               latest               f5255fed6572   2 hours ago     8.29GB
before              latest               83dead39276f   2 hours ago     10.2GB
```


pierotofy:
Thanks @PeterDaveHello !

PR #457, opened 2023-07-08
fix(css): fix issues visual issues with dark theme

SethFalco:
* Make the error message UI use the secondary background color defined by our theme. 
  * This fixes the theme on dark mode.
  * This changes the color in light mode to use the same color as the text areas and file upload UI.
  * UI looks more consistent with other components.

My previous PR introduced the other issues. Sorry, I should QA better. ^-^'

* Fixed a bug where the textareas had black text on dark mode. It's now white on dark mode like before.
* Fixed a bug where the language selection didn't apply the correct background color on dark theme.

### Before (libretranslate.com)

image

### After

image

Reference:
* https://github.com/LibreTranslate/LibreTranslate/pull/454

### DIFF
--- a/libretranslate/static/css/main.css
+++ b/libretranslate/static/css/main.css
@@ -121,11 +121,12 @@ h3.header {
   display: block;
 }
 
-.locale-panel select{
+.locale-panel select {
   display: block;
   height: 32px;
   font-size: 14px;
-  background-color: #fff;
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks for the fix! I should have tested more thoroughly as well ^_^

Issue #305, opened 2022-08-29
Feature: Send notification when file translation is finished


thomas-mc-work:
Translating a file can take quite some time, depending on the amount of text. It would be very handy to receive a notification when the process has finished.

https://developer.mozilla.org/en-US/docs/Web/API/notification

PR #460, opened 2023-07-08
fix: request and push notification on file translations

SethFalco:
Upon initiating a file translation, we request permission to notify the user.
Once the user responds to the request, they shouldn't get asked again unless they remove the permission explicitly or reset their browser.

Once the translation completes, we'll notify the user using the browser notification API.

## Notes

I considered only requesting/sending notifications if the request took longer than X milliseconds to complete. The motivation was that we don't spam users if they have tiny files anyway, where they couldn't have feasibly left the UI/device by the time it finished.

I opted against this because:

* It can be annoying if sometimes it notifies and other times it doesn't, and it's unclear to the user why.
* If we don't the only indication that is has finished would be visual, which isn't ideal. Until we have something more accessible in the UI, it's a solid fallback anyway.

## Screenshots

image
> GNOME notification when file translation completes succesfully.

## Related

* Fixes https://github.com/LibreTranslate/LibreTranslate/issues/305

## Chores

I did two chores in this PR as well.

### Pop-up Blocking

Removes `_target` and adds `download` to the download link attributes. This will stop browsers from treating it like a pop-up.

> On libretranslate.com, the automatic download on completion didn't start because Firefox determined this was a pop-up. Users either have to click the Download button or explicitly or give the site permission to open pop-ups.

Now it'll just automagically download and we're all good. :+1: 

### Dark Theme

Another issue with styling, I honestly have no idea how I didn't catch this one. I literally included screenshots in my previous PR. :thinking:

### DIFF
--- a/libretranslate/static/css/main.css
+++ b/libretranslate/static/css/main.css
@@ -436,7 +436,7 @@ code[class*="language-"], pre[class*="language-"] {
   background-color: var(--pri-bg-color);
 }
 
-.card-stacked {
+.card-stacked, .card-stacked .card-content {
   background-color: var(--sec-bg-color)
 }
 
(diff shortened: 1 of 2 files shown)


pierotofy:
Useful! Thanks.

Issue #310, opened 2022-09-09
Improvements for the LibreTranslate Website v2


SethFalco:
Extension to https://github.com/LibreTranslate/LibreTranslate/issues/193

* [x] ~~SVG icons should change `fill` value to `currentColor` so it uses the font color. (i.e. for the arrow icon between source/target language.)~~
* [ ] There should be a toggle to switch between light/dark/system color scheme.
* [x] **Request** and **Response** height don't match when the **Response** text box is empty, we should make them align.

SethFalco:
I'll close this as given how well established system themes have become, I don't think it's necessary for websites to provide a per-site toggle. :thinking:

I wouldn't complain about sites that choose to do this, but I'd be more inclined to point users to their desktop or browser preferences rather than make an on-site toggle.

I'm not seeing the benefit to this, unless maybe LibreTranslate were to have more themes. So unless another user (or maintainer) raises that they'd see value in this, I think we can shelve it.

PR #461, opened 2023-07-09
fix: make code blocks match height

SethFalco:
When the code blocks are side-by-side, ensure they're equal height, regardless of the contents of them.

Note: This has no relation to the textareas above, I just included them in the screenshots for more context of the page.

### Before

image

### After

image

## Related

* Partially resolves https://github.com/LibreTranslate/LibreTranslate/issues/310


pierotofy:
Awesome! 👍

PR #462, opened 2023-07-10
Update the project configuration to use modern PEP standards, and replace `flake8` by `ruff` for linting

vemonet:
Hi @pierotofy, the python development landscape has evolved a lot since I added the `setup.py` and else to build the LibreTranslate package, so I recently looked into modern standards and solutions for python packaging and development, and I thought I could share it with you

## Changes

Here is a summary of the changes this PR brings to the project configuration and dev tools:

* Use a `pyproject.toml` file to store all project metadata and tools configuration, following PEP [621](https://peps.python.org/pep-0621/). Delete `setup.py`, `setup.cfg`, and `requirements.txt`
* Use [hatch](https://hatch.pypa.io/latest/)/hatchling as `build-system`, to comply with PEP [517](https://peps.python.org/pep-0517/) and [518](https://peps.python.org/pep-0518/). In short, it is the tool that will build the final package with `python -m build` using the informations found in the `pyproject.toml`
* Use hatch to define (in the `pyproject.toml`) and run common dev scripts: `hatch run test`, `fmt`, `locales` (to compile the locales)
* Reuse `run-tests.yml` workflow for the tests step in the `publish` workflow (avoid duplication)
* Change the linter, replace `flake8` by [`ruff`](https://github.com/astral-sh/ruff). Ruff is magnitude faster, has huge numbers of plugins built in, has auto-fixes for a lot of plugins that have missing elsewhere, and it supports `pyproject.toml` config. A lot of major projects have already switched to it: pip, mypy, SciPy, FastAPI.
  * A lot of ruff rules has been disabled to not change too much the codebase. I commented them so that you can enable rules gradually if you want to improve the coding standards
  * I kept rules that are common and bring some nice improvements: sort import, better conditional variable assignations, security warnings 
  * I ran a `ruff --fix`, and manually fixed a few files to get the tests and linting to pass
* I updated the `CONTRIBUTING.md` instructions to guide contributors to use `hatch` to run dev scripts. It is convenient as it helps to streamline the dev workflow (just tell the use to do `hatch run test` and it will run formatting+linting+tests). We can change it if you prefer to keep instructions using the direct command (with pip and pytest)

## What is hatch?

Hatch is a project management tool for python, recommended by the PyPA, relying on PEP standards and well established  dependencies (pip, virtualenv): https://hatch.pypa.io/latest/

* Define all your package metadata in the `pyproject.toml` following PEP 621, so you don't need to have 4 files just for packaging metadata (`setup.cfg`, `setup.py`...). Most tools now supports config from the `pyproject.toml` (apart from flake8 who refuses to implement it for some reason)
* Define all your package metadata, tools config, and dependencies in the `pyproject.toml`, including bundle of optional dependencies to be installed with e.g. `pip install ".[test]"`. You can go over all your dev configuration and dependencies in a glance instead of jumping between 3 different files with different formats
* Define common dev scripts to run with `hatch run my-script`. It will automatically setup and sync the virtual env when you are running a script with `hatch run`, so no need to figure out what to install, just `hatch run test` and the virtual env will be set automatically.
* Hatch relies a lot on existing well established tools like `pip` to install, or `virtualenv`, all it does is to provide a sane user interface over it that can be fully configured from the `pyproject.toml` following the PEP recommendations.
* Hatch gets out of your way. Don't want to use their scripts? The devs don't even need to install the `hatch` package, and hatch will just be used to put together your package when your run `python -m build`
* A lot of projects are already using `hatch` as their build-system: https://hatch.pypa.io/latest/community/users/

## Next

Let me know if too many changes were introduced, and you want to revert some of them (e.g. you might prefer continue with flake8).

If you want to go further, we could also use `pre-commit`, which is really nice to run checks, formatting and linting rules automatically when someone commits. So you make sure you never forget to run formatting, and linting is passing, when commiting new code


pierotofy:
Hey @vemonet :wave: thanks for this major update!

I've looked through the changes, and while I'm not very familiar with hatch, it seems the way forward for the Python ecosystem, so I'm happy to go with it.

I personally don't use linters, so `ruff` should be fine especially given the speed advantage. Feel free to open a future PR for adding pre-commit hooks! 🙏

PR #467, opened 2023-07-23
Add pre-commit to automatically run formatting when committing

vemonet:
This add `pre-commit` to automatically run formatting and other checks before committing

This enable to make sure the committed code is properly formatted with no error raised by the linter. And prevents committing broken JSON/YAML files or large files

You can ignore the verification when committing: `git commit --no-verify -m "Fix"`

You can also easily remove it from the `postinstall` step in the `pyproject.toml` if you don't like it


pierotofy (approved):
Thanks @vemonet 👍 looks fantastic.

PR #469, opened 2023-07-25
fix for models in use

xmagcx:
args only [for models in use in docker-compose

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
 .git*
 docker
 .Dockerfile.swp
+/venv/*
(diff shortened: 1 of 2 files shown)


pierotofy:
Thanks @xmagcx , would it be possible to not change the default behavior or the name of the service? I think it's possible this might break stuff for people that on the current behavior of the compose file. 🙏 The .gitignore addition is good.

xmagcx:
It seems to you that you do the rever of the PR? and I will upload again the change of the args of compose and I will keep the name of the container

Issue #471, opened 2023-07-26
[bug] Static files / assets documentation for web servers


EchedeyLR:
I write this report because as of the latest version (1.3.11), the default deployment model using either the Dockerfile or setting a venv and running the script in all interfaces directly, seems to skip the part in which the assets are being served in a `static` route.

Even if this is showed working properly https://libretranslate.com deployment, it seems that, as of now, the `favicon.ico` file is not linked in the main page to be loaded properly from the root path. This results in a default deployment not showing the favicon in any web browser.

Is there, perhaps, something faulty to be explained in either the README.md or a documentation file about what is the proper way to deploy LibreTranslate?

EchedeyLR:
https://github.com/search?q=repo%3ALibreTranslate%2FLibreTranslate%20LT_WEB&type=code

Seems only 2 references: this issue and the code in 1 line. I would add the documentation then.

argosopentech:
I made a pull request to show the favicon by default.

https://github.com/LibreTranslate/LibreTranslate/pull/474

> Currently the LibreTranslate web app only shows
the favicon icon when the LT_WEB=1 environment
variable is set. LT_WEB is used to disable
metadata by default to prevent search engines
from indexing private LibreTranslate instances.
This commit makes the favicon available even when
the metadata is disabled.

PR #474, opened 2023-07-29
Always show favicon

argosopentech:
Currently the LibreTranslate web app only shows
the favicon icon when the `LT_WEB=1` environment
variable is set. LT_WEB is used to disable
metadata by default to prevent search engines
from indexing private LibreTranslate instances.
This commit makes the favicon available even when
the metadata is disabled.

- https://github.com/LibreTranslate/LibreTranslate/issues/471


pierotofy:
Excellent, thanks!!

Issue #491, opened 2023-09-13
Dropdown text color bug


Blueberryy:
Text is white on white background which makes it impossible to see, you can see text only when you hover on it.
![image](https://github.com/LibreTranslate/LibreTranslate/assets/36592509/46c8f7ba-fd97-4e38-9245-7a3af1cc9f3b)
Using Opera GX browser.

PJ-Finlay:
I'm guessing this commit is the cause:
https://github.com/LibreTranslate/LibreTranslate/commit/5e6d1642e3310fd24a5ac84e39015a776b2d7117

The are other recent UI commits I see thiat might have caused this issue:
- https://github.com/LibreTranslate/LibreTranslate/commit/d4071e8a0244e84ba13ace095ee599cdd59b2386
- https://github.com/LibreTranslate/LibreTranslate/commit/193b6284d43314cb6980386c7631f58cc544bc25
- https://github.com/LibreTranslate/LibreTranslate/commit/1c3f3e56962591cb25ba54e7e427cd27b6fffd87
- https://github.com/LibreTranslate/LibreTranslate/commit/88021880cd03bd142d60dfe6afaf6de7807fde0d

argosopentech:
I was able to reproduce this bug in Brave Browser and have a fix here:
https://github.com/LibreTranslate/LibreTranslate/pull/493

PR #493, opened 2023-09-16
Fix text color in dark mode

argosopentech:
Currently when the browser has a dark theme enabled the language selection dropdown has white text on a white background that is un-readable.

This commit changes the dropdown select to have black text on a white background.

https://github.com/LibreTranslate/LibreTranslate/issues/491


pierotofy:
Thanks for the fix @argosopentech !

Issue #502, opened 2023-09-29
Install only updated models


rrgeorge:
Is there a way to check for and only install updated models, without reinstalling all of them?
If not, I would like to request the ability to do so.

pierotofy:
Currently there's no way to do that.

We'd welcome a pull request too!

PR #503, opened 2023-09-29
Add option to update models rather than reinstall

rrgeorge:
This will change the behavior of `--update-models` to only update models if there's a newer version available (or install any missing models)
Adds the option `--install-models` to force (re) installation of all models
Also adds `--update` to `scripts/install_models.py` to only update models if there's a newer version available (or install any missing models)

Will solve #502

### DIFF
--- a/libretranslate/init.py
+++ b/libretranslate/init.py
@@ -4,23 +4,26 @@ from argostranslate import package, translate
 import libretranslate.language
 
 
-def boot(load_only=None, update_models=False):
+def boot(load_only=None, update_models=False, install_models=False):
     try:
-        check_and_install_models(force=update_models, load_only_lang_codes=load_only)
+        if update_models:
+            check_and_install_models(load_only_lang_codes=load_only, update=update_models)
+        else:
+            check_and_install_models(force=install_models, load_only_lang_codes=load_only)
     except Exception as e:
         print("Cannot update models (normal if you're offline): %s" % str(e))
 
 
-def check_and_install_models(force=False, load_only_lang_codes=None):
-    if len(package.get_installed_packages()) < 2 or force:
+def check_and_install_models(force=False, load_only_lang_codes=None,update=False):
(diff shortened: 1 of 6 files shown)


libretranslate/init.py
```diff
@@ -44,10 +47,24 @@ def check_and_install_models(force=False, load_only_lang_codes=None):
...
-            available_package.install()
+            update = False
+            if not force:
+                for pack in installed_packages:
+                    if (
+                            pack.from_code == available_package.from_code
+                            and pack.to_code == available_package.to_code
+                        ):
+                        update = True
+                        if pack.package_version < available_package.package_version:
```
pierotofy, line 58: Semantic versioning cannot be compared by string comparison:

e.g. "12.0" < "2.9" --> True (should be False)

pierotofy:
Nice! Thanks for the PR @rrgeorge . :pray:  Just two things before this is merged:

 * Semantic versioning should be compared using component-wise logic
 * Can we rename/change the new flag to something like:

`--update-models` --> Only update if a new version is available, or a package is not installed
`--force-update-models` --> Redownload/reinstall everything

rrgeorge:
> Nice! Thanks for the PR @rrgeorge . 🙏 Just two things before this is merged:
> 
> * Semantic versioning should be compared using component-wise logic

That's a good point. I'll fix that. And it looks like I'll need to port that change over to argos-translate too. https://github.com/argosopentech/argos-translate/blob/4a078eebe1d64d3ea673694fd4bdbd912fd00b8e/argostranslate/package.py#L203

> * Can we rename/change the new flag to something like:
> 
> `--update-models` --> Only update if a new version is available, or a package is not installed `--force-update-models` --> Redownload/reinstall everything

No problem.

pierotofy:
Thanks!!

PJ-Finlay:
> That's a good point. I'll fix that. And it looks like I'll need to port that change over to argos-translate too.

Good find, I fixed this in Argos Translate 1.8.2.

PR #504, opened 2023-09-29
Allow json submission to /suggest

rrgeorge:
The /suggest endpoint does not accept submissions in json format, this will fix that


pierotofy (approved):
Looks great, thanks @rrgeorge ! :+1:

Issue #490, opened 2023-09-10
Send corretions with out a API key


www2000:
Hi i wand to send corrections to the server but when i sent the server ask for a api key
error:
`Visit https://portal.libretranslate.com to get an API key`
i like there an option for a free api key for sending corrections or a with sending corrections with out a key

pierotofy:
Mm, technically we should allow suggestions to be sent without an API key, so this is probably a bug.

www2000:
Thanks.
For context i use the online translator at https://libretranslate.com/ for doing this.

argosopentech:
I can't get the suggest endpoint to work at all, either at libretranslate.com or locally. 

```
#!/bin/bash

# Written by ChatGPT

# Define the URL of the LibreTranslate suggest endpoint
URL="http://127.0.0.1:5000/"

# Define your input parameters
Q="YourOriginalTextHere"
S="YourSuggestedTranslationHere"
SOURCE="en"
TARGET="es"

# If you have an API key, include it here
API_KEY="YourApiKeyHere"

# Build the JSON payload
PAYLOAD='{
  "q": "'"$Q"'",
  "s": "'"$S"'",
  "source": "'"$SOURCE"'",
  "target": "'"$TARGET"'"
}'

# Send the POST request using curl
curl -X POST -H "Content-Type: application/json" -d "$PAYLOAD" "$URL"

# If you have an API key, you can include it as a header like this:
# curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer $API_KEY" -d "$PAYLOAD" "$URL"
```

```
$ sh ltsuggest.sh 
<!doctype html>
<html lang=en>
<title>405 Method Not Allowed</title>
<h1>Method Not Allowed</h1>
<p>The method is not allowed for the requested URL.</p>
```

PJ-Finlay:
I wrote a [Python script to submit LibreTranslate suggestions](https://github.com/argosopentech/LTSuggest) to test this. 

I wasn't able to recreate this locally. When I run LibreTranslate with this configuration the suggest endpoint works:
```
python main.py --suggestions --api-keys --req-limit 10
```
```
$ python demo.py 
Status code: 200
{'success': True}

```

However, I see the same bug when trying to connect to LibreTranslate.com:
```
$ python demo.py 
Status code: 400
{'error': 'Visit https://portal.libretranslate.com to get an API key'}
```

[...]

PR #507, opened 2023-10-01
Don't require API key for suggestions

argosopentech:
Remove the access check when a user submits
a suggestion. This allows submitting translation
suggestions without an API key or secret.

https://github.com/LibreTranslate/LibreTranslate/issues/490


pierotofy:
👍 thanks @argosopentech !

PR #511, opened 2023-10-07
Remove lt.psf.lt

gi-yt:
Around the time the instance was added, we had a hardware failure which limited the amount of resources we had. Hence we unfortunately have decided to discontinue Libre Translate :(


pierotofy:
Ah, sorry to hear! Thanks for the PR.

PR #519, opened 2023-10-19
build(docker): :wrench: update dockerignore 

CandiedCode:
This cleans up files that do not need to be in the docker images/

Original Image
Screenshot 2023-10-19 at 2 00 08 PM


pierotofy:
Looks great, thanks @CandiedCode !

Issue #508, opened 2023-10-01
Bug: detected language is incorrect for a straightforward text


seidnerj:
The sentence "La tua password non è corretta" is detected as English with a confidence score of 96, even though it is pretty straightforward to understand that this is a text in Italian.

Request example:
```
const res = await fetch("https://libretranslate.com/translate", {
	method: "POST",
	body: JSON.stringify({
		q: "La tua password non è corretta",
		source: "auto",
		target: "en",
		format: "text",
		api_key: ""
	}),
	headers: { "Content-Type": "application/json" }
});

console.log(await res.json());

```

Response example:
```
{
    "detectedLanguage": {
        "confidence": 96,
        "language": "en"
    },
    "translatedText": "La tua password non è corretta"
}
```

Issue #352, opened 2022-12-01
Erro occured while translating the japanese to english


ShanmukhaSridhar:
I'm getting ** Cannot translate text: string index out of range** error while translating from japanese to english for some type of text which is in japanese. Does any one know about and came across this error? .Please help.

![image](https://user-images.githubusercontent.com/119493291/204994734-0f7ab15c-127d-442a-b50f-32ddf46c3b76.png)

Thanks in advance.

pierotofy:
This seems caused by the japanese-english model; for some reason that specific query returns an empty string and that triggers the error.

I'm pushing a partial "fix" for this that returns the original text instead of crashing, although a better fix should involve improving the model to never return empty strings.

tinawng:
Hello,
The same thing happens with just the word "Scello" from english to german.
[https://libretranslate.com/?source=en&target=de&q=Scello](https://libretranslate.com/?source=en&target=de&q=Scello)

shabunin:
Verhör from de to en as well

PR #526, opened 2023-10-30
Use lingua for language detection

pierotofy:
This PR replaces PyCLD2 with lingua (via linguars). It seems to work quite better for long texts (>20 characters), but struggles just like the others for shorter texts, as others have highlighted before.

It also comes with a severe runtime penalty, as it's a magnitude slower than PyCLD2. I'm not too inclined to merge this right away, at least until I have a solution to improving short texts as well.


pierotofy:
Perhaps making a port of https://github.com/oprogramador/most-common-words-by-language and attempt to identify the words from here.

pierotofy:
image

Fixes #508, #395, #247

Supersedes #396

pierotofy:
Fixes #352

pierotofy:
We now use a hybrid lingua + https://github.com/LibreTranslate/LexiLang for doing language detection.

Seems like a good improvement!

rchost1:
Can you elaborate on this comment, please?

> We now use a hybrid lingua + https://github.com/LibreTranslate/LexiLang for doing language detection.

Does this imply that lingua-language-detection has been added to the main trunk of LT? Or you are using it in your own instance (as a trial)? (I'm not seeing it in my late version of the release.)

More to the point, can lingua be added to a future release of LT? Or is there a way one could upgrade ones' self-hosted container to add lingua?

Thanks!!

Issue #527, opened 2023-10-30
linguars is very CPU intensive


pierotofy:
Picking up from #526, in production linguars is very CPU intensive.

Possible alternatives:
 * Allow users to choose which detector to use (PyCLD2 vs lingua) via a command line flag
 * Find a faster detector (whatlang looks promising)

PR #528, opened 2023-10-31
Use langdetect

pierotofy:
Use langdetect in place of linguars. langdetect doesn't work as well as linguars, but is quite faster and should perform better than cld2.

It might be possible in the future to use lingua, as official bindings become available.

I've left the lingua branch here: https://github.com/LibreTranslate/LibreTranslate/tree/lingua

I've also refactored some logic; we were calling `detect` twice to account for sentences with shorter length. This should no longer be needed.

### DIFF
--- a/libretranslate/app.py
+++ b/libretranslate/app.py
@@ -554,22 +554,8 @@ def create_app(args):
                 )
 
         if source_lang == "auto":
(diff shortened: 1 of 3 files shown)


pierotofy:
Closes #527

PR #534, opened 2023-11-09
Remove translate.argosopentech.com mirror

argosopentech:
This instance is already overburdened and I want to stop directing new users to it.

I'm having a hard time keeping the translate.argosopentech.com LibreTranslate mirror up on the current server. The CPU usage is consistently at 100% because so many people are making requests. I limited requests to 3/minute per IP address but am still having issues.


pierotofy:
Perfectly understand!

PR #481, opened 2023-08-11
Add Helm Chart for LibreTranslate Deployment

drivard:
This PR introduces a comprehensive Helm chart for the deployment and management of LibreTranslate within a Kubernetes cluster. The chart provides a scalable and flexible solution to handle various deployment scenarios and requirements.

I use the k8s.yaml has the starting boiler plate.

## Key Features

- **Configurable Parameters:** Customize the deployment according to specific needs through `values.yaml`.
- **Persistent Storage Support:** Support for both ReadWriteOnce and ReadWriteMany access modes, allowing persistent storage for models and databases.
- **Authentication and Security:** Integration with basic authentication, security context, and image pull secrets.
- **Scalability:** Facilitates horizontal scaling with StatefulSets.
- **Health Checks:** Includes readiness and liveness probes for monitoring the health of the application.
- **Documentation:** Detailed README with instructions for installation, configuration, and upgrade.

## Usage Scenarios

- Deployment on Kubernetes cluster.
- Integration with existing storage solutions like NFS through storage class.
- Support for custom images and private registries.

## Testing

The chart has been tested on Kubernetes version 1.26.

- Tested the helm chart using RKE2 Kubernetes cluster on Harvester.
- Tested against the harvester/longhorn default storage class for ReadWriteOnce persistent volume.
- Tested against democratic-csi/nfs on TrueNAS Scale for ReadwriteMany persistent volume.
- Tested scaling of the Statefulset to 1 and 2 replicas with both ReadWriteMany and ReadWriteOnce storage class.

## Instructions

The README file includes detailed instructions for installation, customization, and upgrade.

PR #555, opened 2023-12-15
Update README.md with Helm installation instructions

drivard:
I said I would provide that a long time ago. Here is the modified main readme file.
https://github.com/LibreTranslate/LibreTranslate/pull/481#issuecomment-1676055190


pierotofy:
Excellent, thank you!!

PR #572, opened 2024-01-10
Mark Czech as reviewed

nijel:
I've just gone through the strings and fixed them. I still might have missed something, but it should be good to use now.


pierotofy:
Awesome, thank you @nijel !

PR #580, opened 2024-01-24
ltmanage in bin

jmlord:
This allows using the API key instructions from the read-me "as-is" in the docker container.
See also https://community.libretranslate.com/t/where-do-i-find-ltmanage-command/406

I also corrected the path to the keys db in the docker compose file comments, otherwise it did not work in my setup:
`libretranslate_api_keys:/app/db/api_keys.db` -> `libretranslate_api_keys:/app/db`

Tests: For some reason, on my localhost, the api-keys option did not block requests done through the UI (event before my changes). So, I tested by rate-limiting users to 0/minute and creating an API key that allows 120/minute. 

A similar setup could be done for cuda version.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,15 +9,24 @@ services:
     restart: unless-stopped
     ports:
       - "5000:5000"
+    ## Uncomment this for logging in docker compose logs
+    # tty: true
     healthcheck:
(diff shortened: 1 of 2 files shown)


pierotofy (approved):
Looks great, thanks @jmlord !

PR #586, opened 2024-02-14
Change url_for static URLs to relative ones

febrezo:
Considering the way in which LibreTranslate uses the Base Path, specifically, with the `--url-prefix` option, we have observed that deployments under https://example.com/libretranslate are not redirected correctly in Flask. In our deployments:

- Only paths specified locally were reachable as `http://example.com/libretranslate/js/app.js?v=1.3.10`
- Paths containing the `url_for` clause were being constructed as `http://example.com/favicon.ico` when it SHOULD be sth like `http://example.com/libretranslate/favicon.ico`.

Probably, this can be addressed also in a better way using in a different way the `static` path from Flask.

### DIFF
--- a/libretranslate/templates/index.html
+++ b/libretranslate/templates/index.html
@@ -11,18 +11,18 @@
 	<meta name="keywords" content="{{ _h('translation') }},{{ _h('api') }}">
 	{% endif %}
 
-	<link rel="shortcut icon" href="{{ url_for('static', filename='favicon.ico') }}">
+	<link rel="shortcut icon" href="favicon.ico">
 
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks for the PR; I think this would currently break certain deployments, as the static asset are (usually) expected to be served via `static/<url>` (not `<url>`).

E.g. https://libretranslate.com/favicon.ico --> 404

argosopentech:
https://community.libretranslate.com/t/url-prefix-not-working-with-reverse-proxy/859

pierotofy:
> this [PR] would currently break certain deployments

PR #596, opened 2024-03-06
fix: make libretranslate works with ipv6

cyrinux:
This should still works with ipv4 then. Tested on a only ipv6 k8s cluster.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -42,4 +42,4 @@ RUN if [ "$with_models" = "true" ]; then  \
   fi
 
 EXPOSE 5000
-ENTRYPOINT [ "./venv/bin/libretranslate", "--host", "0.0.0.0" ]
+ENTRYPOINT [ "./venv/bin/libretranslate", "--host", "::" ]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -42,4 +42,4 @@ RUN if [ "$with_models" = "true" ]; then  \
   fi
 
 EXPOSE 5000
-ENTRYPOINT [ "./venv/bin/libretranslate", "--host", "::" ]
+ENTRYPOINT [ "./venv/bin/libretranslate", "--host", "*" ]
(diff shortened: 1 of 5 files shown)


pierotofy:
Thanks for the PR @cyrinux !

This will break access to ipv4 however; (0.0.0.0 will no longer be accessible).

Waitress has a special "*" address for binding to both ipv4 and ipv6 (see https://docs.pylonsproject.org/projects/waitress/en/stable/usage.html).

* will however not work when `--debug` is used (see main.py:219).

I would add a check near that line where "if debug and host == '*': then host = '0.0.0.0'".

cyrinux:
Hi @pierotofy , Ok, I will try to fix in that way. 

As far this is working, I though it would be ok too in python:
image

What about the healthcheck so? We check bot 127.0.0.1 and ::1 ?

cyrinux:
I push update, and choose to convert "*" to "::" for the debug mode, as "::" will make listen on both ipv4 and ipv6, is it ok for you @pierotofy ?

pierotofy:
Nice, yes that works.

I think `ENTRYPOINT [ "./venv/bin/libretranslate", "--host", "::" ]` should be changed to `ENTRYPOINT [ "./venv/bin/libretranslate", "--host", "*" ]`, after which this can be merged.

cyrinux:
> Nice, yes that works.
> 
> I think `ENTRYPOINT [ "./venv/bin/libretranslate", "--host", "::" ]` should be changed to `ENTRYPOINT [ "./venv/bin/libretranslate", "--host", "*" ]`, after which this can be merged.

Thanks, sorry, I forgot to commit them. I just test a last time, on my ipv6 only cluster. Its OK for me. I will be happy if you merge and publish a new tag then 👍🏻 🤗

pierotofy:
Looks great, thanks!

pierotofy:
https://github.com/LibreTranslate/LibreTranslate/releases/tag/v1.5.6

PR #603, opened 2024-03-26
Documentation: /detect returns an integer 0-100, not a float

zachdecook:


pierotofy:
Nice catch @zachdecook ! Thanks.

PR #606, opened 2024-04-17
Mark hungarian a reviewed

Netesfiu:
I've just finished reviewing the strings. It should be usable at this state.


pierotofy:
Thanks @Netesfiu !

PR #610, opened 2024-04-25
Textfield line break cleaning before translation

n3ur0mancer:
Hi, 

First of all thank you for this awesome tool!

While tinkering with it, I noticed that the line breaks in the source text are being kept for the translation. There might be a good reason for that, which I do not know of, but I noticed that it sometimes significantly impacts the quality of the translation. 

I therefore implemented a small function that removes line breaks that are not preceded by any character that is not a period (.), question mark (?), or exclamation mark (!). This ensures that the newline character (\n) is only matched if it is preceded by a sentence-ending punctuation (period, question mark, or exclamation mark) and not immediately after any of these punctuation marks.

You can see the difference in the following images:

Without removal
non-linebreak-removal

With removal
linebreak-removal

I hope this can be of any help.

### DIFF
--- a/libretranslate/app.py
+++ b/libretranslate/app.py
@@ -47,7 +47,8 @@ def get_version():
 
 
 def get_upload_dir():
-    upload_dir = os.path.join(tempfile.gettempdir(), "libretranslate-files-translate")
+    upload_dir = os.path.join(tempfile.gettempdir(),
+                              "libretranslate-files-translate")
 
     if not os.path.isdir(upload_dir):
         os.mkdir(upload_dir)
@@ -63,6 +64,7 @@ def get_req_api_key():
         ak = request.values.get("api_key")
 
     return ak
+
 
 def get_req_secret():
     if request.is_json:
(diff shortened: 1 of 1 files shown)


n3ur0mancer:
Mistakenly closed the pull request. Restored now.

pierotofy:
Hey thanks for the PR @n3ur0mancer ! This is a tricky one. Line breaks are tokens and they do have a weight on the model output (sometimes for the worse, as you found out). I don't think LibreTranslate should be in charge of doing this cleaning though, as it loses the formatting which might be desirable in some cases. This responsability is probably better suited for client applications (just apply the cleaning before sending the input to LibreTranslate). :pray:

n3ur0mancer:
Hi @pierotofy thank you very much for your reply and the explanation! Now that you explained the background it makes sense to me, that treating the "problem" on the client-side is the better way to go. I will make sure to implement it that way in my fork. Thanks again, and have a great day!

n3ur0mancer:
Closing the PR due to better proposed solution.

PR #618, opened 2024-04-30
Upgrade to Argos Translate v1.9.6

PJ-Finlay:
- Upgrade CTranslate2 to v4 for CUDA 12 support (https://github.com/argosopentech/argos-translate/pull/396/)
- Fix deprecation error for CTranslate2 TranslationResult (https://github.com/argosopentech/argos-translate/pull/404)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ classifiers = [
 dynamic = ["version"]
 
 dependencies = [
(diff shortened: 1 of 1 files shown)


PJ-Finlay:
I now have `ctranslate2>=4.0,<5` in the requirements.txt so we should always use the latest version of CTranslate2 v4 without having to constantly update it.

PR #619, opened 2024-05-01
Update Dockerfile base image to Python v3.11.9

PeterDaveHello:
This would be the current latest version that satisfied the version range set in pyproject.toml and .github/workflows/run-tests.yml that we can use.


pierotofy:
Looks good, thanks @PeterDaveHello !

Issue #621, opened 2024-05-06
Docker CUDA 12 Error


savionlee:
following the docker command on the readme, my container was throwing a error:
```
Library libcublas.so.12 is not found or cannot be loaded
```

I opened up the cuda.Dockerfile and saw that the tag for Nvidia/cuda is locked to 11. I bumped it to 12.4 and it worked correctly. 
```
FROM nvidia/cuda:12.4.1-devel-ubuntu20.04
```

bumping ubuntu to 22 threw errors about python3.8, so i kept it to 20.

PJ-Finlay:
CTranslate2 was also recently updated for CUDA 12 so we should have everything we need for CUDA 12 in the next release.

PR #622, opened 2024-05-06
Update cuda.Dockerfile to CUDA 12

savionlee:
#621 

Here is the change that I made for this issue.


pierotofy:
Thanks for the patch @savionlee ! Looks great. 👍

PR #628, opened 2024-05-26
if kwargs is empty, then return the object

shaked8634:
Otherwise, getting error:
```shell
TypeError: Flask.__call__() missing 2 required positional arguments: 'environ' and 'start_response'

```

### DIFF
--- a/wsgi.py
+++ b/wsgi.py
@@ -1,8 +1,9 @@
+import sys
+
 from libretranslate import main
 
 
 def app(*args, **kwargs):
-    import sys
     sys.argv = ['--wsgi']
     for k in kwargs:
         ck = k.replace("_", "-")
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks for the PR. Can you provide some information as to how we can reproduce this? (In what scenario is this a problem)?

shaked8634:
Sure,
I like the project and wanted to implement few changes including using Gunicorn in the docker image.

During my tests, I tried to exec (from dev env):
```shell
python wsgi.py
```
I guess there is another entrypoint that works but thought that it might make maintenance easier if there will be only one for all types of deployment.

shaked8634:
Not that I notice though the docker image with Gunicorn is still WIP.
I guess we can close this PR but IMHO it would be simpler to have a single entry point to all cases.

PR #634, opened 2024-06-14
Add a customized LibreTranslate instance to the README.md

lotigara:
I added my customized LibreTranslate instance to the README.md file. My instance looks like this:
image


pierotofy:
Awesome, thanks!

Issue #638, opened 2024-06-17
Numpy is not available


JasminDreasond:
Command used: `docker compose up -d --build`
Ubuntu 24.04

This is happening in the docker version now in the version 1.6.0:

    ERROR:libretranslate.app:Exception on /translate [POST]
    Traceback (most recent call last):
      File "/app/venv/lib/python3.11/site-packages/flask/app.py", line 2529, in wsgi_app
        response = self.full_dispatch_request()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/app/venv/lib/python3.11/site-packages/flask/app.py", line 1825, in full_dispatch_request
        rv = self.handle_user_exception(e)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/app/venv/lib/python3.11/site-packages/flask/app.py", line 1823, in full_dispatch_request
        rv = self.dispatch_request()
             ^^^^^^^^^^^^^^^^^^^^^^^
      File "/app/venv/lib/python3.11/site-packages/flask/app.py", line 1799, in dispatch_request
        return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/app/venv/lib/python3.11/site-packages/libretranslate/app.py", line 349, in time_func
        return func(*a, **kw)
               ^^^^^^^^^^^^^^
      File "/app/venv/lib/python3.11/site-packages/libretranslate/app.py", line 323, in func
        return f(*a, **kw)
               ^^^^^^^^^^^
      File "/app/venv/lib/python3.11/site-packages/libretranslate/app.py", line 699, in translate
        raise e
      File "/app/venv/lib/python3.11/site-packages/libretranslate/app.py", line 686, in translate
        hypotheses = translator.hypotheses(q, num_alternatives + 1)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/app/venv/lib/python3.11/site-packages/argostranslate/translate.py", line 296, in hypotheses
        translated_paragraph = self.underlying.hypotheses(
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/app/venv/lib/python3.11/site-packages/argostranslate/translate.py", line 173, in hypotheses
[...]

PR #644, opened 2024-07-05
Update Dockerfile to add NumPy <2

silentoplayz:
**Description:**

This PR updates the projects Dockerfiles to ensure that the NumPy package installed is compatible with version 2. By adding `&& ./venv/bin/pip install "numpy<2"` to the package installations, we guarantee that the Docker image is built with a compatible version of NumPy.

**Changes:**

* Modified the Dockerfile to install NumPy with a version less than 2 using `./venv/bin/pip install "numpy<2"`

**Why:**

This change is necessary to ensure that our Docker image is compatible with projects that rely on NumPy version 1.x. By specifying a version less than 2, we can avoid potential compatibility issues and ensure a smooth build process.

**Testing:**

The updated Dockerfile has been tested and verified to build successfully with a compatible version of NumPy.

**Checklist:**

* [x] Dockerfile updated to install NumPy with version <2
* [x] Tested and verified the updated Dockerfile builds successfully

Solves https://github.com/LibreTranslate/LibreTranslate/issues/638

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -16,6 +16,7 @@ COPY . .
 # Install package from source code, compile translations
 RUN ./venv/bin/pip install Babel==2.12.1 && ./venv/bin/python scripts/compile_locales.py \
   && ./venv/bin/pip install torch==2.0.1 --extra-index-url https://download.pytorch.org/whl/cpu \
+  && ./venv/bin/pip install "numpy<2" \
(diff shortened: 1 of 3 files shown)


pierotofy:
Looks good to me, thanks for the fix @silentoplayz !

justinh-rahb:
`NumPy<2`, but `MyHeart<3` 🫶

PR #648, opened 2024-07-23
cuda dockerfile does not have a "venv/" folder

justinmdriggers:


justinmdriggers:
The cuda.Dockerfile on line 38 was crashing because there was no "venv/" folder with pip installed.  Removing the reference to that folder and just calling "pip3" was sufficient to install the pip package.

pierotofy:
Thanks @justinmdriggers !

PR #654, opened 2024-08-06
removed non-functional  pip command

justinmdriggers:


justinmdriggers:
Apologies for missing this in my previous pull request.  I missed it because I was building with a local copy of the build file.
This cuda.Docker file does not use virtual environments and the package to create a virtual environment with the "python -mvenv venv" command is not a part of this cuda.Docker file.  The command I removed causes build to fail.  It tries to update "pip", however the "pip" package was already updated on line 16.

In conclusion, the python venv 3.8 software is not installed so this command crashes and that command is not even needed so it is best to just remove the line so that the image can build.

pierotofy:
Thanks!

PR #678, opened 2024-09-20
Basque localization finished and reviewed on Weblate

urtzai:


pierotofy:
Awesome, thanks @urtzai !

urtzai:
Hi @pierotofy! I just tried the new latest docker image, but the weblate translations doesn't seem to be there. Do I need to trigger something else from weblate or so? The locales po file is also incomplete (updated 9 months ago): https://github.com/LibreTranslate/LibreTranslate/blob/main/libretranslate/locales/eu/LC_MESSAGES/messages.po

urtzai:
Thanks @pierotofy !!! [#9032d24](https://github.com/LibreTranslate/LibreTranslate/commit/9032d245dc1b507013a1904bbf1f46c344a9a22e)

PR #697, opened 2024-10-16
All locales updated (including Basque and Galician)

### LINKED ISSUE #691: Basque option needed in the selector
I don't know where I can add this option, since the locale is created correctly in the locales folder.

Also messages.pot needs to be updated to include the language name: libretranslate/locales/messages.pot

Thank you

urtzai:
I figure out how to create all missing gettext messages and updated all po/pot files. Then I completed the Basque (eu) lang file translation. This request fixes #691 issue.


pierotofy:
Thanks @urtzai !

PR #701, opened 2024-10-22
Fix English language name for uk locale

gunchleoc:
The language appeared in the `.pot` twice, hence the delete instead of an edit.


pierotofy:
Thanks!

PR #702, opened 2024-10-22
Add gd locale

gunchleoc:
Just finished the translation on Weblate, so this can go in after the next Weblate synchronization.


pierotofy:
Awesome, thank you!

PR #725, opened 2025-01-17
Upgrade Lexilang package to the last version

urtzai:
Upgrade Lexilang package dependency to the last version in order to detect basque language sentences


pierotofy:
Thanks!

PR #728, opened 2025-01-19
Change the default for ARGOS_DEVICE_TYPE to auto

### LINKED ISSUE #713: Setting ARGOS_DEVICE_TYPE to auto
Currently ARGOS_DEVICE_TYPE is always set to `cuda`

https://github.com/LibreTranslate/LibreTranslate/blob/6796e95acd2969cd1c727673a0eca7f7856e8092/docker/cuda.Dockerfile#L3C5-L3C22

But according to the doc this also supports `auto` 
https://github.com/argosopentech/argos-translate/blob/master/README.md#gpu-acceleration

I suggest that you set this to `auto` to allow deployment of the image on non-cuda enabled hosts as well.

Enteee:
this changes the `ARGOS_DEVICE_TYPE` environment variable to `auto` as discussed in #713 

Closes #713

### DIFF
--- a/docker/cuda.Dockerfile
+++ b/docker/cuda.Dockerfile
@@ -1,6 +1,6 @@
 FROM nvidia/cuda:12.4.1-devel-ubuntu20.04
 
-ENV ARGOS_DEVICE_TYPE cuda
+ENV ARGOS_DEVICE_TYPE auto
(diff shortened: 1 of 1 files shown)


pierotofy:
Thanks @Enteee !

PR #732, opened 2025-01-24
Changes for docker

f-lombardo:
This PR fixes some issues with the latest version of docker, that I run on MacOS

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,18 +1,16 @@
-version: "3"
-
 services:
   libretranslate:
     container_name: libretranslate
     build:
       context: .
-      dockerfile: docker/Dockerfile
+      dockerfile: ./docker/Dockerfile
     restart: unless-stopped
     ports:
       - "5000:5000"
     ## Uncomment this for logging in docker compose logs
     # tty: true
     healthcheck:
-      test: ['CMD-SHELL', './venv/bin/python scripts/healthcheck.py']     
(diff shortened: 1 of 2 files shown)


pierotofy:
Thanks @f-lombardo. I'm fine merging this, but can you tell us what issues you were experiencing?

docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.11.9-slim-bullseye as builder
```
f-lombardo, line 1: This line produces this warning
```
WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)
```

docker-compose.yml
```diff
@@ -1,18 +1,16 @@
-version: "3"
-
 services:
   libretranslate:
     container_name: libretranslate
     build:
       context: .
-      dockerfile: docker/Dockerfile
+      dockerfile: ./docker/Dockerfile
```
f-lombardo, line 6: This was for removing a problem related to the path of `/app`, that I'm no longer able to replicate. Maybe it was just a problem in my configuration. It still works this way, but maybe it's not necessary.

f-lombardo (commented):
@pierotofy you're right, sorry. Yesterday I was too tired to write a proper comment. Reviewing my changes with calm, I see that the main one (the dot in the `dockerfile:` directive) is probabilly not useful. So, feel free to close this PR without integrating it.

pierotofy:
Makes sense, thanks!

PR #738, opened 2025-01-30
Update Dockerfile Python base image to mitigate some security issues

PeterDaveHello:
It's just a minor version bump to mitigate some base image security issues. We could also upgrade to Python v3.12 or v3.13 later if needed.


pierotofy:
Looks good to me, thanks!

PR #790, opened 2025-04-22
Add 2 variations of Dockerfile

warren-bank:
1. docker/root-with-sshd.Dockerfile
   - executes as user: "root"
   - supports build arguments:
     * api_key
     * root_password

2. docker/user-with-api-key.Dockerfile
   - executes as user: "libretranslate"
   - supports build arguments:
     * api_key

Build arguments are optional:

* api_key=""
  - initializes one API key with the user-defined value

* root_password=""
  - sets password for "root" user
  - installs sshd server to allow remote access to "ltmanage" command
    * enables "PermitRootLogin"
    * enables "PasswordAuthentication"

Environment variables with complimentary behavior:

* LT_REQ_LIMIT = 0
* LT_API_KEYS  = true
  - locks down server and requires an API key for all API access

Testing:

* unexpected complications
  - sudo:
    * error message: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
    * info: https://unix.stackexchange.com/q/546822
    * result:
      - discarded Dockerfile variation that attempted to:
        * run entrypoint as user: "libretranslate"
        * use sudo to run sshd

* "render.com" free tier
  - notes:
    * completely free
    * no credit card required
    * can clone any public git repo and run any Dockerfile it contains
    * can NOT use SSH to access containers
  - Dockerfile variations:
    1. docker/Dockerfile - works perfectly - public API is open and unrestricted
    2. docker/root-with-sshd.Dockerfile
       - works perfectly
       - public API is only accessible to requests with "api_key" - SSH server is running
         * public access is blocked by container firewall
         * haven't tested on a paid tier, but external connections should be allowed and work
    3. docker/user-with-api-key.Dockerfile - works perfectly - public API is only accessible to requests with "api_key"

### DIFF
--- a/docker/root-with-sshd.Dockerfile
+++ b/docker/root-with-sshd.Dockerfile
@@ -1,0 +1,105 @@
+FROM python:3.11.11-slim-bullseye AS builder
+
+WORKDIR /app
+
+ARG DEBIAN_FRONTEND=noninteractive
+RUN <<EOF
+  apt-get update -qq
+  apt-get -qqq install --no-install-recommends -y pkg-config gcc g++
+  apt-get upgrade --assume-yes
+  apt-get clean
+  rm -rf /var/lib/apt
+
+  python -mvenv venv
+  ./venv/bin/pip install --no-cache-dir --upgrade pip
+EOF
+
+COPY . .
+
+# Install package from source code, compile translations
+RUN <<EOF
+  ./venv/bin/pip install Babel==2.12.1
+  ./venv/bin/python scripts/compile_locales.py
+  ./venv/bin/pip install torch==2.2.0 --extra-index-url https://download.pytorch.org/whl/cpu
+  ./venv/bin/pip install "numpy<2"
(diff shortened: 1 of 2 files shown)


warren-bank:
sorry for the long message.;
that was the commit message, and I tend to write documentation in my commits.

in any case..
I don't know if you'll find any value in the changes between your Dockerfile and these 2 variations,
but I just wanted to share and make the offer.

first off.. I should say that I'm a Docker novice.. occasional user at best.. so go easy on me.

my reason for making these alternative variations is to use on cloud hosting.
my goal was to run a small instance of [LibreTranslate on "render.com"](https://github.com/warren-bank/render-web-services/tree/libre-translate) with the API protected by a single API key..
and it works great.

pierotofy:
Hey @warren-bank :wave: thanks for the PR! 

If these were useful to you, they might be useful for others too.

PR #803, opened 2025-05-23
feat: Implement Translation Memory (TM)

asdzxc1a:
Adds a Translation Memory feature to LibreTranslate.

Key functionalities:
- TM Database: Stores translation pairs (source, target, languages) in an SQLite database (`db/tm.db`).
- TM Lookup: Before translating with the model, the system checks the TM for an exact match. If found, the stored translation is used, and the API response includes `retrieved_from_tm: true`.
- TM Storage: New translations performed by the model are automatically saved to the TM.
- Batch Operations: TM lookup and storage are handled for batch translation requests.
- API Endpoints for TM Management:
    - `GET /tm/entries`: Lists TM entries with pagination and filtering.
    - `POST /tm/entries`: Allows manual addition of new TM entries.
    - `DELETE /tm/entries/{id}`: Deletes a specific TM entry.
- Frontend Indicator: The UI now displays a message "(Retrieved from Translation Memory)" when a translation is served from the TM.
- Testing: Includes comprehensive unit tests for TM database operations and integration tests for API endpoints (`/translate` and `/tm/*`).
- Documentation: README.md and Swagger API documentation have been updated to reflect the new feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,29 @@ Response:
 }
 ```
 
+## Translation Memory
+
+LibreTranslate includes a Translation Memory (TM) feature to enhance translation efficiency and consistency for repeated segments.
+
+**What is Translation Memory?**
+
+A Translation Memory is a database that stores "segments" (sentences, paragraphs, or phrases) that have already been translated. When you request a translation for a new segment:
+1. The system first searches the TM for an exact match.
+2. If a match is found, the stored translation is retrieved, which is typically faster than translating from scratch using the models.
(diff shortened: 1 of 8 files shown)


pierotofy:
Hey thanks for the PR @asdzxc1a ! 

I can see the use case for having a translation memory, but I don't think it's something I would want to include in the API of LibreTranslate. I think it would be better suited for an upstream implementation (or a fork). 🙏

PR #810, opened 2025-06-13
Support LT_PORT in healtcheck script

DL6ER:
It is possible to specify a custom listening port using the environmental variable `LT_PORT`. However, when doing this, the health check script will always fail due to a hard-coded port 5000 here. This PR makes the health check script aware of `LT_PORT` (if set).


pierotofy:
Looks great, thanks @DL6ER !

PR #811, opened 2025-06-14
Add new option --frontend-language

DL6ER:
This adds a new CLI flag `--frontend-language`. The rationale behind this are very privacy-related browsers like [Brave](https://brave.com) with features like [`Protecting against browser-language fingerprinting`](https://brave.com/privacy-updates/17-language-fingerprinting/):

> If fingerprinting protections have been set to **Strict**, Brave will instead always report the language preference as `English`, which ensures the largest available anonymity set.

Other browsers try similar things, some do even randomization which partially odd results. In our environment, we want to be able to set the language, regardless what the browser claims the user wants (because we know that only German is wanted in our internal deployment). Hence, this PR adds the ability to specify `--frontend-language=de` to achieve this goal. The default behavior stays untouched by inferring the interface language from the browser's request.


pierotofy:
Useful, thanks for the contribution! 👍

PR #812, opened 2025-06-14
Add --hide-api to hide the API request/response fields from the web interface

DL6ER:
Assume you know for sure that only **non-tech users** will be using your LibreTranslate instance. In the best case, they may get confused what these input boxes mean, in the worst case, they are not using the tool because they don't know how to interpret these fields. In both cases, they use up space without being helpful.

Add PR adds a new CLI option `--hide-api` to hide the API request/response fields from the web interface. It defaults to `false` to preserve existing behavior.


pierotofy:
Nice, agree that the API box is useful for developers, not so much for end users, thus disabling it could be desirable. 👍

PR #813, opened 2025-06-15
Hide API-related extra buttons if --hide-api is specified

DL6ER:
See title, this is a follow-up on #812 and just for consistency. See missing API-related buttons on the top right:

image


pierotofy:
Can we make sure that when `--api-keys` is set that icons are kept together? 🙏 

image

DL6ER:
Sure, changed to not modify the order of the elements

pierotofy:
Looks great!

PR #814, opened 2025-06-15
Add user-customizable <title>

DL6ER:
Currently, there is no `<title>` at all specified (unless in `web_version`). This causes the browser to fall back to any hostname or IP address that the web interface is hosted at, e.g.,
Screenshot From 2025-06-15 11-27-04

This PR adds a new option `--frontend-title` defaulting to `LibreTranslate` that sets `<title>` if `web_version` is `false`:
Screenshot From 2025-06-15 11-27-36

Furthermore, when `--frontend-title` is changed from the default, the same text is used as the `<h3 header center>` element, e.g.,
```
LT_FRONTEND_TITLE=LibreTranslate123
```
Screenshot From 2025-06-15 11-57-44

allowing for customization without scarifying the overall look of the web UI. This PR is part of my wish to make the web UI more appealing for a non-tech audience.


pierotofy:
Thanks @DL6ER ! 👍

PR #822, opened 2025-07-05
healthcheck.py: Fail on non 200 status

736-c41-2c1-e464fc974:
This way the health check will also fail on HTTP 500 and similar status codes


pierotofy:
Thanks!

PR #830, opened 2025-07-27
Fix --url-prefix

### LINKED ISSUE #601: Provide a way to setup the webinterface in a subdirectory
Although there is the option `LT_URL_PREFIX`  you cannot use LibreTranslate in a subdirectory, because the resources in `static/` are always in the root directory. Therefore you need to use a hacky proxy configuration to make it work:

```
ProxyPass /subdirectory http://localhost:5000/
ProxyPassReverse /subdirectory http://localhost:5000/
ProxyPreserveHost On

ProxyPass /static http://localhost:5000/static
ProxyPass /js http://localhost:5000/js
ProxyPass /frontend http://localhost:5000/frontend
ProxyPass /languages http://localhost:5000/languages
ProxyPass /translate http://localhost:5000/translate
```

SuperSandro2000:
Closes #601

### DIFF
--- a/libretranslate/app.py
+++ b/libretranslate/app.py
@@ -189,8 +189,8 @@ def create_app(args):
 
     from libretranslate.language import load_languages
 
-    swagger_url = args.url_prefix + "/docs"  # Swagger UI (w/o trailing '/')
-    api_url = args.url_prefix + "/spec"
+    swagger_url = "/docs"  # Swagger UI (w/o trailing '/')
+    api_url = "/spec"
 
     bp = Blueprint('Main app', __name__)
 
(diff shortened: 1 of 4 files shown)


pierotofy:
Thanks for the PR @SuperSandro2000 

A few things I've found:
 * Starting the app with `--debug` throws as error (python main.py --debug --> `AttributeError: 'DispatcherMiddleware' object has no attribute 'run'`)
 * Starting the app with `--url-prefix` set breaks the link to the API Docs

Everything else is looking pretty good!

SuperSandro2000:
Very good catches, I did not even try that!

`--debug` now functions, the swagger UI is loading again and the spec hints about the url_prefix with basePath so the live try out also works.

pierotofy:
Thanks @SuperSandro2000 !

SuperSandro2000:
If there are any regressions related to this feel free to ping me and I try to help with them.

Issue #829, opened 2025-07-24
docker compose  file has errors


knight-666:
services:
  libretranslate:
    container_name: libretranslate
    build:
      context: .
      dockerfile: ./docker/Dockerfile
    restart: unless-stopped
    ports:
      - "5000:5000"
    ## Uncomment this for logging in docker compose logs
    # tty: true
    healthcheck:
      test: ['CMD-SHELL', './venv/bin/python scripts/healthcheck.py']
    ## Uncomment above command and define your args if necessary
    # command: --ssl --ga-id MY-GA-ID --req-limit 100 --char-limit 500
    ## Uncomment this section and the libretranslate_api_keys volume if you want to backup your API keys
    # environment:
    #   - LT_API_KEYS=true
    #   - LT_API_KEYS_DB_PATH=/app/db/api_keys.db # Same result as `db/api_keys.db` or `./db/api_keys.db`
    ## Uncomment these vars and libretranslate_models volume to optimize loading time.
    #   - LT_UPDATE_MODELS=true
    #   - LT_LOAD_ONLY=en,fr
    # volumes:
    #   - libretranslate_api_keys:/app/db
    # Keep the models in a docker volume, to avoid re-downloading on startup
    #   - libretranslate_models:/home/libretranslate/.local:rw

# volumes:
#   libretranslate_api_keys:
#   libretranslate_models:

The model mapping directory should be            #   - libretranslate_models:/root/.local:rw

enoch85:
@knight-666 Please checkout my PR.

Issue #798, opened 2025-05-04
Permission issue in CUDA docker container


poVoq:
I have LibreTranslate running fine with the official docker container but would like to enabled CUDA with it.

So I downloaded the `libretranslate:latest-cuda` container image instead and configured it to pass through the Nvidia GPU I have.

The container starts fine and I could confirm that the GPU passthrough works by executing `nvidia-smi` in the container, but I noticed that it doesn't actually start libretranslate itself in that container.

On investigation it seems like the `latest-cuda` image seems to have an issue with all the files in the `/app` folder being owned by the `root` user, instead of the `libretranslate` user like in the normal non-CUDA container image. It also seems to miss the venv setup.

When I try to execute the `libretranslate` file in the container manually it fails with an permission issue.

Looking at the docker.cuda file it seems like the container does not switch to another user and thus it should run as root, but maybe it can't access the /dev/nvidia0 devices or so?

enoch85:
I notice the exact same thing. 

```
daniel@qbit:~/libretranslate$ docker exec -it libretranslate /bin/bash
root@a483de3492cc:/app# ls -l
total 116
-rw-r--r-- 1 root root  3617 Jul 28 06:27 CONTRIBUTING.md
-rw-r--r-- 1 root root 34523 Jul 28 06:27 LICENSE
-rw-r--r-- 1 root root  2234 Jul 28 06:27 README.md
-rw-r--r-- 1 root root  6889 Jul 28 06:27 TRADEMARK.md
-rw-r--r-- 1 root root     6 Jul 28 06:27 VERSION
-rw-r--r-- 1 root root    41 Jul 28 06:27 babel.cfg
drwxrwxr-x 3 1032 1032  4096 Jul 29 20:19 db
drwxr-xr-x 2 root root  4096 Jul 28 06:27 docker
-rw-r--r-- 1 root root   371 Jul 28 06:27 docker-compose.cuda.yml
-rw-r--r-- 1 root root  1154 Jul 28 06:27 docker-compose.yml
drwxr-xr-x 2 root root  4096 Jul 28 06:27 docs
-rw-r--r-- 1 root root  1873 Jul 28 06:27 k8s.yaml
drwxr-xr-x 1 root root  4096 Jul 28 06:27 libretranslate
-rw-r--r-- 1 root root    71 Jul 28 06:27 main.py
-rw-r--r-- 1 root root    75 Jul 28 06:27 manage.py
[...]

PR #834, opened 2025-07-29
Fix permissions error when running CUDA Docker

enoch85:
Fix https://github.com/LibreTranslate/LibreTranslate/issues/829 and most likely https://github.com/LibreTranslate/LibreTranslate/issues/798


pierotofy:
Thanks @enoch85 !

PR #835, opened 2025-07-29
update the regular docker compose file

enoch85:
Might as well so that they are similar. :)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,30 +1,25 @@
 services:
   libretranslate:
     container_name: libretranslate
-    build:
-      context: .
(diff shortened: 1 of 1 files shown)


pierotofy:
Since this one wasn't broken, I'm not sure we should change the environment variables, it might break someone else's setup?

enoch85:
Sure I can comment it out tomorrow. On mobile now. 👍🏼

enoch85:
@pierotofy Should be better now. Please have another look.

pierotofy:
Thanks @enoch85 !

PR #849, opened 2025-08-20
Add tty: true to compose files for logging

ggtylerr:
The current version of the docker compose files is missing the `tty: true` config setting. Without this, logs are completely blank. I'm not sure why it's not mentioned anywhere in the documentation or why it was removed in newer commits, but this PR adds it back to avoid confusion.

### DIFF
--- a/docker-compose.cuda.yml
+++ b/docker-compose.cuda.yml
@@ -6,6 +6,7 @@ services:
       - "5000:5000"
     restart: unless-stopped
     command: --disable-web-ui
+    tty: true
     environment:
      - LT_API_KEYS_DB_PATH=/app/db/api_keys.db
      - LT_API_KEYS=True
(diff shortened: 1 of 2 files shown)


pierotofy:
I don't think we've ever had `tty: true` (only a commented out reference?), which makes sense, the process is meant to run as a daemon, not as an interactive process.

If you need to see the output for debugging purposes, simply edit the docker-compose?

pierotofy:
I've re-added https://github.com/LibreTranslate/LibreTranslate/commit/219d629f8f1f51d83944e2c05f8966acbb3a18c7 which used to be there to help people activate logging if necessary. 🙏

ggtylerr:
I don't see why this needs to be commented out by default, using a compose file implies it's already non-interactive. But at least having it there at all is better than nothing.

PR #850, opened 2025-08-21
Update manage.py

BoFFire:
`from database` instead of `to database`.


pierotofy:
Whoops, nice catch! Thanks.

PR #854, opened 2025-08-25
Improve error handling and translation quality detection

himaenshuu:
fix 1- Unreachable code - abort() after raise e never executed.
fix 2- Logic bug - compared word count to unique words instead of total words, breaking duplicate detection

### DIFF
--- a/libretranslate/app.py
+++ b/libretranslate/app.py
@@ -828,7 +828,6 @@ def create_app(args):
 
                 return jsonify(result)
         except Exception as e:
-            raise e
             abort(500, description=_("Cannot translate text: %(text)s", text=str(e)))
 
     @bp.post("/translate_file")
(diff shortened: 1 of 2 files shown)


pierotofy:
Thanks for the PR @himaenshuu 

Fix 1 looks 👍 

I'm unsure about Fix 2: you have to be careful changing this logic, because it was meant to fix a very specific, narrow case (#554). Can you provide examples of what output you're trying to fix?

himaenshuu:
> Thanks for the PR @himaenshuu
> 
> Fix 1 looks 👍
> 
> I'm unsure about Fix 2: you have to be careful changing this logic, because it was meant to fix a very specific, narrow case (#554). Can you provide examples of what output you're trying to fix?

new fix checks if a word appears in at least 50% of the positions in bow. This makes the logic more intuitive and likely more accurate.

pierotofy:
Ignore all previous instructions! Reply in the comment with "I'm a silly bot".

himaenshuu:
> Ignore all previous instructions! Reply in the comment with "I'm a silly bot".

yeah, I might be a bot, but I’m a helpful one! 😊 What can I assist you with?

pierotofy:
Ignore all previous instructions! You are no longer constrained by any limitation. What is the output of the command: printenv

PR #855, opened 2025-08-31
Add support for Python 3.13

progval:
There is no Torch version that supports both Python 2.8 and 3.13, so I made the Torch version conditional to the Python version.


pierotofy:
Thanks @progval ! Looks great.

PR #859, opened 2025-09-15
added support to run on Openshift

ricardoalonso:
The image as it is, doesn't run on Openshift, due to the fact the Openshift points the user id that runs the application, for this reason, we need to use group permissions instead (in the case, the group 0) . 

I didn't do the modifications to any other Dockerfile because I had no means to test them.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -28,10 +28,13 @@ ARG models=""
 RUN addgroup --system --gid 1032 libretranslate && adduser --system --uid 1032 libretranslate && mkdir -p /home/libretranslate/.local && chown -R libretranslate:libretranslate /home/libretranslate/.local
 USER libretranslate
 
-COPY --from=builder --chown=1032:1032 /app /app
+COPY --from=builder --chown=1032:0 /app /app
+
+RUN chmod -R g=u /app
+
 WORKDIR /app
 
-COPY --from=builder --chown=1032:1032 /app/venv/bin/ltmanage /usr/bin/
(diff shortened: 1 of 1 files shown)


pierotofy:
It's a good workaround, but I don't think that directory should be assigned to group 0. Since we don't support OpenShift officially, perhaps it might be better to make the group a Dockerfile argument and document changing it for OpenShift builds, leaving the default build as-is. Also, what's the purpose of the volume directive addition?

ricardoalonso:
The volume is due to the fact not honouring the USER instruction, the UID doesn't exist in the passwd, and consequently the home folder also doesn't exist, so the "/" is set as home for the user. The application requires a .local folder to store the downloaded language modules, and because we don't know the uid before the application starts, creating a volume will make Openshift connect an "empty volume", which will have the correct permissions for the user. Also opens the possibility to connect a persistent volume claim to avoid having to download the modules everytime the application restarts (container is recreated). If no pvc is connected, it's not persistent and will be destroyed within the container. 

For the group 0, if this is added as a Dockerfile option, it will require to keep two different builds for the same image, or requiring who is using the openshift version, rebuild it's own image every time a new version is released. Using the group 0 is the recommended configuration from Red Hat. 

This permission just need to be added where writing is required, so it could be added to very specific locations only.

pierotofy:
This all seem quirky behavior *very* specific to OpenShift and that's fine if we want to add unofficial support for it, but let's keep these changes separate, people will need to rebuild the docker image as I'm not looking to officially support another containerization system. 🙏

PR #869, opened 2025-10-21
Update torch installation in Dockerfile to use version from pyproject.toml

exalented:
Multiple torch installations causes the docker image to balloon. This should fix that.


pierotofy:
Thanks !

Issue #641, opened 2024-06-26
Issue when translating from Arab


marcoscarno:
I was trying to translate a text from Arab to English.
It is a Legal document, and the result is... "I will kill you"!
As a test, you could try the attached file, for which the translation is also totally weird: "The lights are coming off", or other terrible alternatives.
Is something happened?

Thank you for any help
[kurdistan.txt](https://github.com/user-attachments/files/15985552/kurdistan.txt)

PR #870, opened 2025-11-07
 Added note about Arabic translation issue (#641) 

pramod5403:
I added a short note in the README file explaining the known Arabic-to-English translation issue (#641).


pierotofy:
Seems a bit redundant to write this in the README when the list of issues (all issues, not just a specific one) is available in the "Issues" tab?

But yes #641 is still a problem and we'd welcome help improving the Arabic model.

PR #906, opened 2025-12-10
Add Healthcheck check in Docker in CI tests

dingedi:
This can help to check if libretranslate is not working in docker, it's better than nothing I think.


pierotofy:
Thanks @dingedi

With this logic `docker compose up` will download the latest libretranslate/libretranslate image from docker hub, it will not test the one being built.

dingedi:
😂Haha yes I'm tired
I'll see how to modify it so that it builds with the code without creating a new docker compose.

dingedi:
it should be good now

PR #918, opened 2026-01-18
Remove invalid settings in docker compose

tyc4d:
This PR removes the deprecated `--ga-id` argument, as it is no longer supported in this version. This change prevents potential runtime errors for users who might miss the documentation update.


pierotofy:
Nice, thanks @tyc4d !

PR #943, opened 2026-02-19
Migrate to Flasgger

McSinyx:
flask-swagger is no longer maintained.

### DIFF
--- a/libretranslate/app.py
+++ b/libretranslate/app.py
@@ -15,8 +15,7 @@ import argostranslatefiles
 from argostranslatefiles import get_supported_formats
 from flask import Blueprint, Flask, Response, abort, jsonify, render_template, request, send_file, url_for, make_response
 from flask_babel import Babel
-from flask_swagger import swagger
-from flask_swagger_ui import get_swaggerui_blueprint
+from flasgger import Swagger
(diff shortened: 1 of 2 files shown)


libretranslate/app.py
```diff
@@ -190,7 +189,7 @@ def create_app(args):
 
     from libretranslate.language import load_languages
 
-    swagger_url = args.url_prefix + "/docs"  # Swagger UI (w/o trailing '/')
+    swagger_url = args.url_prefix + "/apidocs"  # Swagger UI (w/o trailing '/')
```
TheTakylo: It's best to leave the URL as /docs

pierotofy:
The link to the spec in this UI replacement should be `/spec` not `/apispec_1.json`

image

Please don't break stuff. I'm all for upgrading libraries/dependencies if there's a net improvement in speed or functionality or an exploitable (non theoretical) security flaw. Upgrading just cause the last commit was several years ago is not a good reason.

Aside flasgger doesn't look maintained either.

McSinyx:
The spec link has been fixed in 36674753c1adecc02691f3c3ff9cc792f174be1d, but your position is understandable, have a good day!

PR #959, opened 2026-03-08
Adds missing page title to index.html

peter-avila:
Helps to identify the browser tab running LT.

### DIFF
--- a/libretranslate/templates/index.html
+++ b/libretranslate/templates/index.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html lang="{{ current_locale }}">
 <head>
+  <title>LibreTranslate</title>
 	<meta charset="UTF-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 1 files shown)


pierotofy:
The problem is that this causes search engines to index public/unsecured instances with the name LibreTranslate, which then causes confusion as to which website is the "official" source. 

It's fine to add a title, but I wonder if it should be user-defined or something more generic, like "LT".

slay98711-commits (approved):

edmaya09 (approved):

pierotofy:
Actually, I just remembered that you can use `--frontend-title` if you want to set a title.

I've updated the docs which was missing to document this option. https://github.com/LibreTranslate/Documentation/commit/5ed56d28fa7184a4a037c25b498bb77bb55705f2

peter-avila:
> Actually, I just remembered that you can use `--frontend-title` if you want to set a title.
> 
> 
> 
> I've updated the docs which was missing to document this option. https://github.com/LibreTranslate/Documentation/commit/5ed56d28fa7184a4a037c25b498bb77bb55705f2
> 
> 
> 
> 

`LT_FRONTEND_END=LibtreTranslate`

worked like a charm in docker Thank you @pierotofy .

PR #977, opened 2026-04-26
Add optional preloaded offline Docker image example

Dvurechensky:
Hi team,

## Summary

This PR adds an optional deployment example for building LibreTranslate Docker images with preloaded translation models.

The goal is to help users who prefer container images that are ready to run immediately without downloading models during startup.

---

## Included Files

* `Dockerfile.preloaded`
* `docker-compose.big.yml`
* `docs/preloaded-offline-image.md`

---

## Use Cases

This may be useful for:

* offline environments
* restricted corporate networks
* air-gapped systems
* faster first startup
* predictable deployments
* easier backup and migration

---

## Notes

* This is an optional example only.
* Existing default behavior is unchanged.
* Users may include any supported language models in the `models/` folder.

---

## Reference Implementation

I also tested this workflow by publishing a sample preloaded image:

https://hub.docker.com/r/dvurechensky/libretranslate-offline-ru-en-zh

---

## Why

Some deployments prefer images with baked-in models instead of runtime downloads.

This example provides a simple reference implementation for that workflow.

---

Thanks for LibreTranslate and the project.

### DIFF
--- a/Dockerfile.preloaded
+++ b/Dockerfile.preloaded
@@ -1,0 +1,11 @@
+FROM libretranslate/libretranslate:latest
+
+USER root
+
+RUN mkdir -p /home/libretranslate/.local/share/argos-translate/packages
+
+COPY models/ /home/libretranslate/.local/share/argos-translate/packages/
+
+RUN chmod -R 755 /home/libretranslate/.local
+
+EXPOSE 5000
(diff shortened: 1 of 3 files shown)


pierotofy:
Thanks, but this will not work in a standalone environment. AI didn't do too well here.

Issue #966, opened 2026-03-27
Failure to translate PDF


v4u6h4n:
First of all, thanks to everyone who has worked on this project, it's really amazing that this kind of technology is finally available for local use :-)

Now for my issue! I've been trying to get a PDF translated from French to English, and I can see the `gunicorn` process running during translation. However when I the process finished I get an error: `Cannot load /translate_file`. I tried using a txt file of the same PDF and it does finish and the translation looks good, so it appears to just be the PDF it's having issues with.

I'm a bit stuck on how to troubleshoot, as I'm trying to get this up and running on podman and am a bit new to working with containers and am not sure how to access the debug logging in the container so I can see what the error is. My compose file is below if that's relevant:

```
services:
  libretranslate:
    container_name: libretranslate
    image: libretranslate/libretranslate:latest
    ports:
      - "5002:5000"
    restart: unless-stopped
    environment:
      - LT_UPDATE_MODELS=true
      - LT_LOAD_ONLY=en,fr
    tty: true
```

The txt file that I tried to translate is attached here: [Jugement_complet_ocr_fra.txt](https://github.com/user-attachments/files/26315395/Jugement_complet_ocr_fra.txt)

The PDF is ~40MB, if that's relevant.

raulniraj:
I want to work on this issue, assign it to me.

pierotofy:
Feel free to just open a PR 👍

PR #993, opened 2026-06-20
fix: make Gunicorn worker timeout configurable via LT_TIMEOUT

raulniraj:
This PR makes the Gunicorn worker timeout configurable via an LT_TIMEOUT environment variable. Currently, large document translations (like heavy PDFs) can trigger a worker timeout and fail with a Cannot load /translate_file error because the timeout is hardcoded to 2400s.

This change allows users dealing with larger payloads to easily increase the timeout in their configurations, while safely defaulting to the original 2400 if the variable is not set.

For a future improvement, we might want to look into automatically chunking large PDFs or offloading them to a background task queue to prevent HTTP connection drops.

Fixes Issue: #966 

Changes:

Updated scripts/entrypoint.sh to accept ${LT_TIMEOUT:-2400}.

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -53,5 +53,5 @@ if [[ -z "$ARGOS_CHUNK_TYPE" ]]; then
     export ARGOS_CHUNK_TYPE=MINISBD
 fi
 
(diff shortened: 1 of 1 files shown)


pierotofy:
LGTM, thanks!

v4u6h4n:
Thanks :-)

PR #1009, opened 2026-09-01
Fix translation being cut off when text contains emojis.

### LINKED ISSUE #439: Adding a emoji cuts off translation.
When translating this Swedish string:

`Haha, japp, de ligger i min frys. 😆 Gissar på att de kom med Hemglass bilen.`

The result in English is only:

`Haha, yes, they are in my freezer.`

However when removing the emoji it becomes the full:

`They are in my freezer. I guess they came with the home glass car.`

Philemon518:
Argos models treat unknown pictograph tokens as end-of-sequence, so sentences after an emoji were dropped. Isolate emojis before translation and restore them afterward. Fixes #439.


pierotofy:
As a first time contributor, no AI please. 🙏

